PR #421, opened 2023-03-07
docs: Update querying-json-rpc.mdx with working code

jbowen93:
Small modifications to golang example to have working example when interacting with a local node via JSON-RPC.

Previous code had following issues:
- Didn't set a context on `client.HeaderByNumber` resulting in `net/http: nil Context` error. 
- Used `header.Number.ToString()` when correct method is `header.Number.String()`. 
- Didn't pass in `blockNumber` to `fmt.Println()` function

Nit: I prefer when I can copy paste example code and have it run so I hardcoded the RPC Endpoint for a local node

Tested by running local `polard` node on M1 Mac following instructions from `main` branch README.md

### DIFF
--- a/docs/web/pages/docs/interacting-with-node/querying-json-rpc.mdx
+++ b/docs/web/pages/docs/interacting-with-node/querying-json-rpc.mdx
@@ -46,18 +46,19 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


itsdevbear:
thx @jbowen93 we are going to add a ton more docs over the week :)

PR #430, opened 2023-03-07
perf(Cosmos Store): DuoCacheKV Store for optimized snapshottable multistores

calbera:


itsdevbear:
yeah makes sense to me, hacky but resolves the case

@yihuang

itsdevbear:
In theory you need a lock around the double read and double write.

yihuang:
> yeah makes sense to me, hacky but resolves the case
> 
> @yihuang

I'm not sure if i fully get the idea, so you write into both `sot` and revision store at the same time, and when commit you can simply keep the dirty states in `sot` store without writing the dirty states layer by layer, is that the idea? but when revert, do you need to undo the changes in `sot` store?
In current PR, the `sot` store is never assigned, so I'm not sure if my imagined implementation is correct.

itsdevbear:
PR isn't done maybe, but Sot (source of truth) is basically a replicates the current top of the cachekv wrap stack without wraps. 

So on finalize we can discard the journal and just write SoT to disk.

itsdevbear:
Hacky af, we were just playing around today

yihuang:
> PR isn't done maybe, but Sot (source of truth) is basically a replicates the current top of the cachekv wrap stack without wraps.
> 
> So on finalize we can discard the journal and just write SoT to disk.

yeah, but when revert, you need to undo the changes in SoT state?

calbera:
you need the revisions because they are the "checkpoints". So if we are currently on snapshot 5, and we revertToSnapshot(2), we apply revision at index 2 to sot. This is probably the most expensive operation of the whole process, but if we assume Reverts happen less often, it may be okay

calbera:
the PR is very incomplete, still missing the RevertToSnapshot logic and Finalize logic (although finalize is very simple)

yihuang:
> you need the revisions because they are the "checkpoints". So if we are currently on snapshot 5, and we revertToSnapshot(2), we apply revision at index 2 to sot. This is probably the most expensive operation of the whole process, but if we assume Reverts happen less often, it may be okay

that sounds like bad worst-case complexity though, and you need to consider the attacker scenario who will exploit the worst case inefficiency.

calbera:
My initial thinking was if we are reverting to cachekv at index `5` , we iterate through the dirty keys in cachekv `5` and set each one of those keys to the corresponding values. We can delete the other keys in sot.

yihuang:
> My initial thinking was if we are reverting to cachekv at index `5` , we iterate through the dirty keys in cachekv `5` and set each one of those keys to the corresponding values. We can delete the other keys in sot.

that won't be a complete state revert, right? you need to consider the changes happen before and after version `5`.

calbera:
True, in theory we would have to replay changes from index 0 to 5 on `sot`. Will pause work on this PR for now and resume if we find a more robust approach.

PR #442, opened 2023-03-09
refactor(cosmos): move rpc and proto into cosmos/

calbera:


calbera:
@itsdevbear can you pull locally and run proto gen

itsdevbear (changes requested):
- need to delete proto from the root folder.
- need to update scripts

itsdevbear:
@calbera 

image

itsdevbear (approved):

PR #443, opened 2023-03-09
feat(plugins): Make historical data an optional plugin

calbera:
- allow block plugin to set headers to store 
- make the historical data plugin storage optional 
- update caching to handle optional historical data


calbera:
still need to fix tests

eth/core/chain_reader.go
```diff
@@ -137,8 +142,13 @@ func (bc *blockchain) GetTransaction(
 			txLookupEntry.BlockNum, txLookupEntry.TxIndex, nil
 	}
 
-	// check the block plugin
-	txLookupEntry, err := bc.host.GetBlockPlugin().GetTransactionByHash(txHash)
+	// check if historical plugin is supported by host chain
+	if bc.hp == nil {
+		return nil, common.Hash{}, 0, 0, ErrTxNotFound
```
itsdevbear, line 147: do we want to return a more specific error that is saying that the historical plugin isnt setup?

calbera: I think saying the historical plugin is oddly specific. From the perspective of getting a tx, you either find it or you don't

itsdevbear (approved):
utACK, also we probably should update the docs + that diagram as well.

Issue #422, opened 2023-03-07
feat: Enable EIP-1153.


itsdevbear:
Should be easy, just wire up TransStore prefixed by contract.

oofncubed:
i can take this

PR #457, opened 2023-03-14
feat(eip-1153): add transient storage

oofncubed:
https://github.com/berachain/polaris/issues/422
smonkey said there would be crayons
image


eth/core/state/statedb.go
```diff
@@ -177,14 +181,19 @@ func (sdb *stateDB) SlotInAccessList(addr common.Address, slot common.Hash) (boo
...
 func (sdb *stateDB) GetTransientState(addr common.Address, key common.Hash) common.Hash {
-	panic("not supported by Polaris")
+	return sdb.transientStorage.Get(addr, key)
 }
 
-// TODO: `SetTransientState` implements the `PolarisStateDB` interface by setting the transient state
+// `GetTransientState` implements `stateDB`
 func (sdb *stateDB) SetTransientState(addr common.Address, key, value common.Hash) {
-	panic("not supported by Polaris")
+	prev := sdb.GetTransientState(addr, key)
```
itsdevbear, line 191: Might be better to just always set? Optimize for the more frequent case (writing to empty slot)

itsdevbear (changes requested):

eth/core/state/interfaces.go
```diff
@@ -123,4 +123,12 @@ type (
 		// `ContainsAddress` returns whether the given `address` is in the access list.
 		ContainsAddress(address common.Address) bool
 	}
+
+	TransientJournal interface {
+		// `TransientJournal` implements `libtypes.Controllable`.
```
calbera, line 128: needs the `libtypes.Controllable[string]` interface here

calbera:
@oofncubed thanks for making the PR!

The idea behind the journals in the StateDB (like `LogsJournal`, etc) is to handle the state itself and not just the "change entries" like its done in the journal in geth. This would get rid of the `transientStorage` field in the stateDB object and offload that logic to the transient journal. And the entries in the stack of each journal is the actual state; in the case of transient storage, it would be the actual map as an entry in the stack.

Let me know if this makes sense.

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,105 @@
...
+	return storage
+}
+
+// `transient` is a `Store` that tracks the transient state.
+type transient struct {
+	ds.Stack[transientStorage]
+}
+
+// `NewTransientStorage` returns a new `transient` journal.
+func NewTransient() *transient {
```
calbera, line 68:
```suggestion
func NewTransientStorage() *transientStorage {
```

eth/core/state/interfaces.go
```diff
@@ -123,4 +123,13 @@ type (
 		// `ContainsAddress` returns whether the given `address` is in the access list.
 		ContainsAddress(address common.Address) bool
 	}
+
+	TransientJournal interface {
```
calbera, line 127:
```suggestion
	TransientStorageJournal interface {
```

This is more accurate and specific in my opinion.

eth/core/state/statedb.go
```diff
@@ -214,6 +218,8 @@ func (sdb *stateDB) Prepare(rules params.Rules, sender, coinbase common.Address,
 			sdb.AddAddress(coinbase)
 		}
 	}
+
+	sdb.TransientJournal.Finalize()
```
calbera, line 222: don't need this here. All of the journals are registered by the controller, which invokes Snapshot, RevertToSnapshot, and Finalize on each of the journals when necessary!

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,105 @@
...
+func (t *transient) AddTransient(account common.Address, key common.Hash, val common.Hash) {
+	currentState := t.Peek().Copy()
+	currentState.Set(account, key, val)
+	t.Push(currentState)
+}
+
+// `GetTransient` returns previous transient storage state for a given account + key.
+func (t *transient) GetTransient(account common.Address, key common.Hash) common.Hash {
+	return t.Peek().Get(account, key)
+}
```
calbera, line 90: The idea of embedding the TransientStorageJournal is that we no longer need to define these methods on the StateDB. These function signatures should be the ones exactly required by the StateDB interface:
`SetTransientState(addr common.Address, key, value common.Hash)`
`GetTransientState(addr common.Address, key common.Hash) common.Hash`

eth/core/state/statedb_test.go
```diff
@@ -116,4 +116,19 @@ var _ = Describe("StateDB", func() {
...
+		value2 := common.Hash{0x03}
+
+		sdb.SetTransientState(addr, key, value)
+		Expect(sdb.GetTransientState(addr, key), value)
+		before := sdb.Snapshot()
+		sdb.SetTransientState(addr, key, value2)
+		Expect(sdb.GetTransientState(addr, key), value2)
+		sdb.RevertToSnapshot(before)
+		Expect(sdb.GetTransientState(addr, key), value)
+	})
```
calbera, line 133: would recommend adding tests for the copy on the journal. Specifically, modifying the head entry on the journal should not modify the (head - 1)th entry on the journal

calbera: also these tests can be in the journal package itself once this logic is removed from the statedb

eth/core/state/interfaces.go
oofncubed: I agree & also renamed the previous transietStorage to be transietState since that felt more accurate

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,105 @@
...
+	t[addr][key] = value
+}
+
+// Get gets the transient storage state for `key` at the given `addr`.
+func (t transientState) Get(addr common.Address, key common.Hash) common.Hash {
+	val, ok := t[addr]
+	if !ok {
+		return common.Hash{}
+	}
+	return val[key]
```
calbera, line 46: this will return `nil` if the `key` doesnt exist for the `addr`, should it return `common.Hash{}` instead? I'm not sure of the desired behavior here.

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,105 @@
...
+		valDeepCopy := make(map[common.Hash]common.Hash, len(storVal))
+		for key, val := range storVal {
+			valDeepCopy[key] = val
+		}
+		storage[storKey] = valDeepCopy
+	}
+	return storage
+}
+
+// `transientStorage` is a `Store` that tracks the transient state.
```
calbera, line 62:
```suggestion
// `transientStorage` is a journal that tracks the transient state.
```

eth/core/state/statedb.go
```diff
@@ -73,14 +75,16 @@ func NewStateDB(sp Plugin) vm.PolarisStateDB {
...
-		LogsJournal:       lj,
-		RefundJournal:     rj,
-		AccessListJournal: aj,
-		ctrl:              ctrl,
-		suicides:          make([]common.Address, 1), // very rare to suicide, so we alloc 1 slot.
+		Plugin:            		  sp,
+		LogsJournal:       		  lj,
+		RefundJournal:     		  rj,
+		AccessListJournal: 		  aj,
+		TransientStorageJournal:  tj,
```
calbera, line 85: fix formatting, spacing off

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,105 @@
...
+		stack.New[transientState](initCapacity),
+	}
+}
+
+// `RegistryKey` implements `libtypes.Registrable`.
+func (t *transientStorage) RegistryKey() string {
+	return transientRegistryKey
+}
+
+
```
calbera, line 79: nitpicky but 2 line spaces is weird. try to keep it to 1 line everywhere

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,105 @@
...
+// `AddTransient` adds a transient change to the `transient` store.
+func (t *transientStorage) SetTransientState(addr common.Address, key, value common.Hash) {
```
calbera, line 81: taking from geth, an obvious optimization here is to only proceed with the copy, set, and push to journal if the given `value` is different from the current value of storage of `addr` at `key`

eth/core/state/journal/transient_storage.go
calbera: yup the one place its used in geth:
image
this will panic if nil returned, which we want to avoid

eth/core/state/journal/transient_storage.go
oofncubed: isn't this conflicting with @itsdevbear 's earlier comment?
https://github.com/berachain/polaris/pull/457#discussion_r1135554559

oofncubed: I think given that writes now require a deepcopy, checking before setting is probably* worth it

itsdevbear: I mean I'm fine either way. You'd have to benchmark the difference tbh

itsdevbear: I'd say just mirror geth if you can

calbera: yeah exactly, makes sense to check before copying. Also runtime of Peek is O(1) so its not expensive there.

eth/core/state/journal/transient_storage.go
oofncubed: I'm also unsure of the desired behaviour here but I guess
pros:
- its fairly explicit that nil is returned only when key doesn't exist for addr, but addr does exist
- consistent with geth
cons:
- its a bit of a gotcha

eth/core/state/journal/transient_storage.go
itsdevbear: yeah true say, forgot it's not going to disk to might as well check facts

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,112 @@
...
+func (t transientState) Get(addr common.Address, key common.Hash) common.Hash {
+	val, ok := t[addr]
+	if !ok {
+		return common.Hash{}
+	}
+	hash, ok := val[key]
+	if !ok {
+		return common.Hash{}
+	}
+	return hash
```
calbera, line 50: lmaooo mb here. you can revert these changes and just return `val[key]`. I realized go maps with a value of an array (in this case `common.Hash`) will actually return the empty array and not nil. nil is only returned when the value is a pointer/interface.
https://go.dev/play/p/IqH6gRQn9sY

oofncubed: its okay, i also missed that lmao

eth/core/state/journal/transient_storage.go
```diff
@@ -0,0 +1,108 @@
...
+	return storage
+}
+
+// `transientStorage` is a journal that tracks the transient state.
+type transientStorage struct {
+	ds.Stack[transientState]
+}
+
+// `NewTransientStorage` returns a new `transient` journal.
+func NewTransientStorage() *transientStorage {
```
calbera, line 68: gonna have to add a `nolint:revive` here. Check the pattern used in logs.go, refund.go, etc

eth/core/state/journal/transient_storage_tests.go
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: BUSL-1.1
```
calbera, line 1: convention is to use filename_test.go, just gotta remove the `s` at the end

calbera:
you're gonna want to run `mage format` from the root of the repo to fix the linter.

calbera (approved):
utACK

itsdevbear:
@ocnc add a test for finalization then lgtm

itsdevbear (approved):
WAGMI 🥸

itsdevbear:
Thanks @oofncubed appreciate the effort king

PR #463, opened 2023-03-14
perf(cosmos): Parallelize converting Cosmos events to Eth logs

calbera:


itsdevbear:
@calbera closing this for now, feels low prio.

PR #470, opened 2023-03-14
fix(statedb): Use a journal for the suicides tracking

calbera:
- Also refactoring statedb and plugin and accesslist journal


itsdevbear:
@calbera douvle check the yellow paper on suicide enforcement

itsdevbear (approved):

PR #476, opened 2023-03-16
feat(precompiles): Distribution Keeper Precompile

neverDefined:
Features:

- Allows users to change wirthdraw addresses 
- Allows users to withdraw staking rewards


itsdevbear:
@neverDefined looks fire G, I'd just say convert the input to the constructor to an interface so that we use our custom distribution module for bera, otherwise looks fire utACK

contracts/src/cosmos/precompile/distribution.sol
```diff
@@ -1,34 +1,59 @@
-// SPDX-License-Identifier: MIT
```
itsdevbear, line 1: This can stay MIT now

itsdevbear (commented):
One more comment

itsdevbear:
@neverDefined if you could get this in today, would be fire.

cosmos/runtime/app.go
```diff
@@ -305,6 +305,8 @@ func NewPolarisApp( //nolint: funlen // from sdk.
 	// THE "DEPINJECT IS CAUSING PROBLEMS" SECTION
 	// ===============================================================
 
+	// Pntr to a Pntr
+	dk := &app.DistrKeeper
```
calbera, line 309: don't need the double pointer here because the keeper's are already initialized by this point

cosmos/runtime/app.go
```diff
@@ -314,7 +316,7 @@ func NewPolarisApp( //nolint: funlen // from sdk.
 			stakingprecompile.NewPrecompileContract(app.StakingKeeper),
 			bankprecompile.NewPrecompileContract(),
 			addressprecompile.NewPrecompileContract(),
-			distrprecompile.NewPrecompileContract(),
+			distrprecompile.NewPrecompileContract(&dk),
```
calbera, line 319:
```suggestion
			distrprecompile.NewPrecompileContract(&app.DistrKeeper),
```

contracts/src/cosmos/precompile/distribution.sol
```diff
@@ -1,34 +1,59 @@
...
+import {IBankModule} from "./bank.sol";
+
 /**
  * @dev Interface of the distribution module's precompiled contract
  */
 interface IDistributionModule {
+    /**
+     * @dev The caller (msg.sender) can set the address that will receive the deligation rewards.
+     */
+    function setWithdrawAddress(address withdrawAddress) external;
```
calbera, line 32: As mentioned in [this issue](https://github.com/berachain/polaris/issues/491), we need to return a boolean value here.

contracts/src/cosmos/precompile/distribution.sol
```diff
@@ -1,34 +1,59 @@
...
+
+    /**
+     * @dev The caller (msg.sender) can set the address that will receive the deligation rewards.
+     * Howver taking in a bech32 address.
+     */
+    function setWithdrawAddress(string calldata withdrawAddress) external;
```
calbera, line 38: here as well

cosmos/precompile/distribution/distribution.go
```diff
@@ -21,30 +21,147 @@
...
 	coreprecompile "pkg.berachain.dev/polaris/eth/core/precompile"
+	"pkg.berachain.dev/polaris/lib/utils"
 )
 
 // Contract is the precompile contract for the distribution module.
 type Contract struct {
 	precompile.BaseContract
+
+	msgServer distributiontypes.MsgServer
+	querier   distributiontypes.QueryServer
```
calbera, line 46: pass these in rather than initializing in the constructor.

contracts/src/cosmos/precompile/distribution.sol
neverDefined: done

cosmos/precompile/distribution/distribution.go
```diff
@@ -21,30 +21,147 @@
...
 }
 
-// NewPrecompileContract returns a new instance of the bank precompile contract.
-func NewPrecompileContract() coreprecompile.StatefulImpl {
+// `NewPrecompileContract` returns a new instance of the bank precompile contract.
+func NewPrecompileContract(dk **distrkeeper.Keeper) coreprecompile.StatefulImpl {
```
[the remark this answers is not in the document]
neverDefined: done

itsdevbear:
@neverDefined whats the status here?

PR #479, opened 2023-03-17
feat(x/evm): Token Middleware

itsdevbear:


cosmos/x/erc20/keeper/middleware.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// Get the ERC20 token corresponding to a given denom.
+	if _, err := k.DenomKVStore(ctx).GetAddressForDenom(denom); err != nil {
+		// If the denom is found, we need to burn the corresponding bank denom.
+		if err = k.BurnCoinsFromAddress(ctx, sender, denom, amount); err != nil {
+			return false, err
+		}
+		// If the token was found, that means we don't have to deploy it. So we need to unlock the lock.
+		k.deployLock.Unlock()
```
itsdevbear, line 75: Defer above if, possible deadlock

itsdevbear:
The more I thing on this, this is such an ugly solution, we should just bite the bullet and do the Create and Call from the precompile. 

I think it's important to show it can be done in an safe way.

itsdevbear:
Having the smart contract perform logic goes against the ethos of why we even built Polaris in the first place.

itsdevbear:
though @smsunarto I'd love your thoughts here. 

Right now we don't allow precompile to call back into the EVM. This means that the ERC20 precompile can't call transfer or transferFrom on a token to move it from someone's wallet after they approve the ERC20 module address, so we have to do it this way.

itsdevbear:
If we were to allow it, the precompile could just call CREATE or CREATE2 to deploy the PolarisERC20 and then CALL to transfer, mint and burn tokens. 

Which feels more correct, though slightly more complex. 

The shim contract feels gross to me.

contracts/src/cosmos/x/erc20/ModuleRouter.sol
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    /**
+     * @dev Transfer tokens to Cosmos
+     * @param denom The denom to transfer.
+     * @param recipient The address of the recipient
+     * @param amount The amount of tokens to transfer
+     */
+    function transferToCosmos(string memory denom, address recipient, uint256 amount) public {
```
smsunarto, line 75: nit, lets make the ordering consistent.

`denom, amount, recipient`

contracts/src/cosmos/x/erc20/ModuleRouter.sol
```diff
@@ -0,0 +1,119 @@
...
+     * @param _erc20Module The address of the ERC20Module
+     */
+    constructor(IERC20Module _erc20Module) {
+        erc20Module = _erc20Module;
+    }
+
+    /**
+     * @dev Transfer tokens to Cosmos
+     * @param token The address of the token to transfer
+     * @param amount The amount of tokens to transfer
```
smsunarto, line 52: nit, the position of amount/recipient is switched.

itsdevbear:
@calbera we gonna nuke the solidity ya?

calbera:
yee

itsdevbear:
closing in favour of: https://github.com/berachain/polaris/pull/524

PR #498, opened 2023-03-20
feat(precompile): Add EVM access to stateful precompiles

calbera:


cosmos/precompile/governance/governance.go
```diff
@@ -96,6 +96,7 @@ func (c *Contract) PrecompileMethods() coreprecompile.Methods {
 // SubmitProposal is the method for the `submitProposal` method of the governance precompile contract.
 func (c *Contract) SubmitProposal(
 	ctx context.Context,
+	_ coreprecompile.EVM,
```
itsdevbear, line 99: Why not use core.PrecompileEVM?

itsdevbear (commented):
Lgtm I think coreprecompile.EVM is kinda ugly, but other than that it's vibes.

itsdevbear (commented):
Lgtm I think coreprecompile.EVM is kinda ugly, but other than that it's vibes.

cosmos/precompile/governance/governance.go
calbera: in eth folder, i just imported `PrecompileEVM` from Geth into the precompile package. `coreprecompile` is meh but had to do that alias because theres a precompile import from cosmos folder

calbera:
we can call it `ethprecompile`

itsdevbear:
yea down that works

itsdevbear (approved):

PR #504, opened 2023-03-21
fix(x/evm): get EVMDenom from store instead of hardcode to "abera"

BrickBera:
tested with "mage start"


calbera:
looks great!

itsdevbear:
utACK 🎉

itsdevbear (approved):

calbera:
can we also remove `"abera"` from other places in the code?

BrickBera:
> can we also remove `"abera"` from other places in the code?

Other places in cosmos/x:
1. https://github.com/berachain/polaris/blob/main/cosmos/x/evm/types/params.go#L32
This is for DefaultParams, should be ok?
2. https://github.com/berachain/polaris/blob/main/cosmos/x/evm/plugins/state/plugin_test.go#L496
This is a test, should be good

itsdevbear:
> 

yeah I think these two are fine. The only thing we could do is abstract it into a variable but I think its not that srs atm

itsdevbear:
gj @BrickBera 🎉

PR #505, opened 2023-03-21
refactor(x/evm): Break host object out of keeper and cleanup.

BrickBera:


cosmos/x/evm/keeper/host/host.go
```diff
@@ -0,0 +1,138 @@
...
+	ethTxMempool sdkmempool.Mempool,
+	offChainKv *offchain.Store,
+	rpcProvider evmrpc.Provider,
+) *Host {
+	// We setup the host with some Cosmos standard sauce.
+	h := &Host{}
+
+	// Build the Plugins
+	h.bp = block.NewPlugin(storeKey)
+	h.Cp = configuration.NewPlugin(storeKey)
```
BrickBera, line 76: have to make this public

cosmos/x/evm/keeper/msg_server.go
```diff
@@ -71,7 +71,7 @@ func (k *Keeper) UpdateParams(
 	}
 
 	// Update the params.
-	k.cp.Prepare(ctx)
-	k.cp.SetParams(&req.Params)
+	k.Host.Cp.Prepare(ctx)
```
BrickBera, line 74: Cp become public because of this usage

cosmos/x/evm/keeper/host/host.go
itsdevbear: Is there a way we can keep it private?

cosmos/x/evm/keeper/msg_server.go
itsdevbear: I think it would be nice if we could keep it private

cosmos/x/evm/keeper/host.go
```diff
@@ -0,0 +1,138 @@
...
+	// We setup the host with some Cosmos standard sauce.
+	h := &Host{}
+
+	// Build the Plugins
+	h.bp = block.NewPlugin(storeKey)
+	h.cp = configuration.NewPlugin(storeKey)
+	h.gp = gas.NewPlugin()
+	h.hp = historical.NewPlugin(h.bp, offChainKv, storeKey)
+	h.txp = txpool.NewPlugin(h.cp, rpcProvider, utils.MustGetAs[*mempool.EthTxPool](ethTxMempool))
+	return h
```
itsdevbear, line 80: return &Host{
	bp: block.NewPlugin(....),
	....
}

cosmos/x/evm/keeper/host.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// GetTxPoolPlugin returns the txpool plugin.
+func (h *Host) GetTxPoolPlugin() core.TxPoolPlugin {
+	return h.txp
+}
+
+// GetAllPlugins returns all the plugins.
+func (h *Host) GetAllPlugins() []plugins.BaseCosmosPolaris {
+	return []plugins.BaseCosmosPolaris{h.hp, h.cp, h.gp, h.pp, h.sp}
```
itsdevbear, line 137: missing some plugins here (I think this was from before your pr tho), also "hp" should be "bp"

cosmos/x/evm/keeper/keeper.go
```diff
@@ -63,15 +50,8 @@ type Keeper struct {
...
-	// The various plugins that are are used to implement `core.PolarisHostChain`.
-	bp  block.Plugin
-	cp  configuration.Plugin
-	gp  gas.Plugin
-	hp  historical.Plugin
-	pp  precompile.Plugin
-	sp  state.Plugin
-	txp txpool.Plugin
+	// The host contains various plugins that are are used to implement `core.PolarisHostChain`.
+	Host *Host
```
itsdevbear, line 54: we can probably make Host private tbh, both the variable and the type.

cosmos/x/evm/keeper/processor_test.go
```diff
@@ -142,9 +142,9 @@ var _ = Describe("Processor", func() {
 			tx := coretypes.MustSignNewTx(key, signer, legacyTxData)
 			addr, err := signer.Sender(tx)
 			Expect(err).ToNot(HaveOccurred())
-			k.GetStatePlugin().CreateAccount(addr)
-			k.GetStatePlugin().AddBalance(addr, big.NewInt(1000000000))
-			k.GetStatePlugin().Finalize()
+			k.Host.GetStatePlugin().CreateAccount(addr)
+			k.Host.GetStatePlugin().AddBalance(addr, big.NewInt(1000000000))
```
itsdevbear, line 146: yeah make host private and provide an accessor.

cosmos/x/evm/keeper/host.go
BrickBera: Hard to do so, because `h.hp = historical.NewPlugin(h.bp, offChainKv, storeKey)` requires h.bp and `h.txp = txpool.NewPlugin(h.cp, rpcProvider, utils.MustGetAs[*mempool.EthTxPool](ethTxMempool))` requires h.cp

cosmos/x/evm/keeper/host.go
```diff
@@ -46,6 +46,18 @@ import (
 // Compile-time interface assertion.
 var _ core.PolarisHostChain = (*host)(nil)
 
+type Host interface {
+	core.PolarisHostChain
+	GetAllPlugins() []plugins.BaseCosmosPolaris
```
itsdevbear, line 51: can we add comments as to why we export these?

cosmos/x/evm/plugins/configuration/plugin.go
```diff
@@ -37,8 +37,8 @@ import (
 
 // Plugin is the interface that must be implemented by the plugin.
 type Plugin interface {
-	plugins.BaseCosmosPolaris
 	core.ConfigurationPlugin
+	plugins.BaseCosmosPolaris
```
itsdevbear, line 41: revert this change, this was me my bad :D

cosmos/x/evm/keeper/host.go
itsdevbear: thats true, good point.

PR #510, opened 2023-03-22
refactor(tests): Make integration tests run more efficiently.

itsdevbear:


magefiles/cosmos.go
```diff
@@ -143,11 +143,11 @@ func (Cosmos) Install() error {
 
 // Runs all main tests.
 func (c Cosmos) Test() error {
-	if err := testUnit(c.directory()); err != nil {
+	if err := TestUnit(); err != nil {
 		return err
 	}
 
-	if err := testIntegration(c.directory()); err != nil {
+	if err := TestIntegration(); err != nil {
```
transmissions12, line 150: adding the ability to run different focussed tests with filter might be powerful here, for the future

transmissions12 (approved):
looks good, bless up

magefiles/cosmos.go
itsdevbear: yuh I agree.

PR #517, opened 2023-03-26
feat: Add more features to testing framework.

itsdevbear:


cosmos/testing/integration/fixture.go
```diff
@@ -36,6 +41,7 @@ const defaultTimeout = 10 * time.Second
 // TestFixture is a testing fixture that can be used to test the
 // Ethereum JSON-RPC API.
 type TestFixture struct {
+	t           network.TestingT
```
transmissions12, line 44: why did we add t here? maybe update the name?

transmissions12 (approved):
looks good, we can add more utility functions as we go

itsdevbear:
So that we can use t.fatal, t is a very common name for this variable

itsdevbear:
Trying to make it so that the fixture is not opinionated to Ginkgo, eventually we can make a test suite that chain unopionated as well.

PR #521, opened 2023-03-27
feat(eip1559): Implement EIP-1559 

itsdevbear:
AYO tbt to when a whole cosmos sdk module was required for this.


calbera:
nice nice, I actually was abt to push up a pr for this

cosmos/x/evm/keeper/processor_test.go
```diff
@@ -139,11 +139,12 @@ var _ = Describe("Processor", func() {
 
 		It("should successfully deploy a valid contract and call it", func() {
 			legacyTxData.Data = common.FromHex(bindings.SolmateERC20Bin)
+			legacyTxData.GasPrice = big.NewInt(10000000000)
```
BrickBera, line 142: intended to be 10 zeros? or should it be 9? idk

PR #527, opened 2023-03-29
fix(precompile): Correctly return Execution Reverted error when precompile errors

calbera:
Using geth update, where error check happens via `errors.Is`, which accounts for wrapping errors.


itsdevbear (approved):
Fire

itsdevbear:
Fix lint then yeet it

PR #546, opened 2023-04-12
feat(eth): Use geth.node for RPC stuff

itsdevbear:
Reuse geths node struct in order to prevent the host chain needing to manage anything rpc related, kinda fire, still a WIP.

Cosmos side affect is that:
- `http://localhost:1317/eth/rpc` -> `http://localhost:8545`
- `ws://localhost:1317/eth/rpc/ws`  -> `ws://localhost:8546` 

But like I'm fine with it. Less stuff for the host chain to have to worry about is bullish.


itsdevbear:
TODO: Cleanup config and setup stuff.

itsdevbear:
The biggest annoyance on the setup front is that the ChainConfig isn't ready until InitGenesis and that the TxPoolPlugin needs a clientCtx, which forces us to do this weird dance of waiting until shit is ready before sending it down into the keeper.

cosmos/precompile/auth/auth.go
```diff
@@ -32,10 +32,11 @@ import (
 	"pkg.berachain.dev/polaris/cosmos/precompile"
 	"pkg.berachain.dev/polaris/eth/common"
 	ethprecompile "pkg.berachain.dev/polaris/eth/core/precompile"
-	"pkg.berachain.dev/polaris/eth/params"
 	"pkg.berachain.dev/polaris/lib/utils"
 )
 
+const requiredGas = 5000
```
calbera, line 38: arbitrary? should probably cost around 1 kvstore read/write?

cosmos/runtime/app.go
```diff
@@ -459,8 +459,7 @@ func (app *PolarisApp) RegisterAPIRoutes(apiSvr *api.Server, apiConfig config.AP
 	if err := server.RegisterSwaggerAPI(apiSvr.ClientCtx, apiSvr.Router, apiConfig.Swagger); err != nil {
 		panic(err)
 	}
-	// Register Ethereum JSON-RPC API as needed by polaris.
-	rpc.RegisterJSONRPCServer(apiSvr.ClientCtx, apiSvr.Router, app.EVMKeeper.GetRPCProvider())
+	app.EVMKeeper.SetClientCtx(apiSvr.ClientCtx)
```
calbera, line 462: nice

eth/provider/config.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/BurntSushi/toml"
+
+	"github.com/ethereum/go-ethereum/node"
+
+	"pkg.berachain.dev/polaris/eth/rpc"
+)
+
+// DefaultConfig returns the default configuration for the provider.
+func DefaultConfig() *Config {
```
BrickBera, line 35: 👍

PR #551, opened 2023-04-17
feat(precompile): governance integration tests and update

neverDefined:


itsdevbear:
Don't we worry it's gonna be too hard to submit gov proposals? 

Can we even write a solidity lib to do this?

PR #558, opened 2023-04-17
fix(test): Fix number of denoms in bank int test + formatting

BrickBera:
fix(test): Fix number of denoms in bank int test + formatting


itsdevbear:
needs lint then lgtm

itsdevbear (approved):

PR #559, opened 2023-04-18
feat(cosmos): Create polaris.toml 

BrickBera:


cosmos/runtime/init.sh
```diff
@@ -67,6 +68,10 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
     	# Set moniker and chain-id (Moniker can be anything, chain-id must be an integer)
 	./bin/polard init $MONIKER -o --chain-id $CHAINID --home "$HOMEDIR"
 
+	cp ./cosmos/runtime/localnode/config/app.toml "$APP_TOML"
```
BrickBera, line 71: don't know if this is good. it works.

config.toml
itsdevbear, line 1: Duplicate file?

cosmos/runtime/init.sh
itsdevbear: Ya works I guess

cosmos/runtime/app.go
```diff
@@ -310,7 +311,17 @@ func NewPolarisApp( //nolint: funlen // from sdk.
 
 	// Polaris Configuration
 	cfg := provider.DefaultConfig()
-	cfg.NodeConfig.DataDir = DefaultNodeHome + "/data"
+	homePath, ok := appOpts.Get(flags.FlagHome).(string)
```
itsdevbear, line 314: This isn't the spot for this. 


I would probably do it in eth/provider and then we should pass the filepath down thru.

itsdevbear:
Also lint

cosmos/runtime/app.go
```diff
@@ -309,8 +310,9 @@ func NewPolarisApp( //nolint: funlen // from sdk.
 	// ===============================================================
 
 	// Polaris Configuration
-	cfg := provider.DefaultConfig()
-	cfg.NodeConfig.DataDir = DefaultNodeHome + "/data"
+	homePath := GetHomePath(appOpts)
```
itsdevbear, line 313: Like pass the base string down to the provider and have the provider do all the parsing and stuff

cosmos/runtime/init.sh
```diff
@@ -84,25 +89,6 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
 	jq '.app_state["mint"]["params"]["mint_denom"]="abera"' "$GENESIS" >"$TMP_GENESIS" && mv "$TMP_GENESIS" "$GENESIS"
```
itsdevbear, line 89: You can delete all this no?

BrickBera: failed to do so. genesis.json has `genesis_time` and account addresses generated

cosmos/runtime/init.sh
itsdevbear, line 1: i'd move init.sh into `localnode` folder.

eth/provider/provider.go
```diff
@@ -38,6 +38,27 @@ type PolarisProvider struct {
...
+) *PolarisProvider {
+	// Load the config file.
+	cfg, err := LoadConfigFromFilePath(configPath)
+	if err != nil {
+		// todo: this is hood.
+		cfg = DefaultConfig()
+	}
+
+	// set the data dir
+	cfg.NodeConfig.DataDir = dataDir
```
itsdevbear, line 54: you don't need this anymore, since you can just put the dataDir in the config?

BrickBera: I don't fully understand, but if I don't put this, it will set DataDir as "/" then we have the folder permission issue

PR #565, opened 2023-04-19
fix(erc20): Generate salts using a nonce from disk

calbera:


cosmos/lib/salt.go
```diff
@@ -0,0 +1,68 @@
...
+	storetypes "cosmossdk.io/store/types"
+)
+
+const globalNonce = `globalNonce`
+
+var globalNonceKey = []byte(globalNonce)
+
+// UniqueDeterministicSalt returns a unique and deterministic salt for the given input bytes. Uses
+// sha256 to hash the input bytes and the global nonce and returns the salt as a *uint256.Int.
+func UniqueDeterminsticSalt(nonceStore storetypes.BasicKVStore, input []byte) *uint256.Int {
```
itsdevbear, line 39: Ask ND he does a similar thing for TokenContainer.

itsdevbear: don't need this whole lib thing.

cosmos/x/erc20/keeper/keeper.go
```diff
@@ -47,15 +47,15 @@ func NewKeeper(
 	authority sdk.AccAddress,
 ) *Keeper {
 	return &Keeper{
-		storeKey:   storeKey,
+		StoreKey:   storeKey,
```
itsdevbear, line 50: this should be private.

cosmos/lib/salt.go
calbera: thought its good practice to do it as a library so even ND can use this

cosmos/x/erc20/keeper/keeper.go
calbera: can make it private but have to regenerate everytime

cosmos/lib/salt.go
itsdevbear: yeah fair, though I don't see why we need to use a salt and CREATE2 when we can just use regular CREATE

calbera: CREATE2 so that an erc20 with the same name doesn't have the same address

PR #350, opened 2023-03-01
feat(noncepool): Tx Noncer

itsdevbear:

PR #571, opened 2023-04-21
feat(txpool): Nonce cache

calbera:
similar, but simpler impl of #350


itsdevbear:
@davidterpay just for ref

itsdevbear (commented):
pushed some changes, it works? I think we need somemore unit testing on this one though, im not confident in the logic

calbera:
current version is the logic used by geth’s light client txpool. now have to support pending and queued txs from geths core txpool

itsdevbear:
Light client doesn't support queued

cosmos/x/evm/plugins/txpool/mempool/mempool.go
```diff
@@ -116,20 +116,20 @@ func (etp *EthTxPool) GetTransactions() coretypes.Transactions {
 
 // GetNonce returns the nonce for the given address from the mempool if the address has sent a tx
 // in the mempool.
-func (etp *EthTxPool) GetNonce(addr common.Address) (uint64, bool) {
+func (etp *EthTxPool) GetNonce(addr common.Address) (uint64, error) {
```
itsdevbear, line 119: I think this might be a regression, since you're going to be returning an error if the nonce is not found, which should just mean nonce 0

calbera: true it should be that simple

calbera:
image
Also adding support for the .Error check, I think we should do this (as Geth does) whenever accessing anything that could have caused an error (nonce, balance specifically)

calbera:
Also we are now exposing the state plugin to be read from multiple threads, so do we need to lock there?

calbera:
This PR should just introduce read-only changes, namely enabling RPC methods: `GetPoolNonce`, `SubscribeNewTxsEvent`

eth/core/chain_reader.go
```diff
@@ -187,7 +187,7 @@ func (bc *blockchain) GetPolarisBlockByNumber(number int64) (*types.Block, error
 	return nil, ErrBlockNotFound
 }
 
-// GetBlockByHash retrieves a block from the database by hash, caching it if found.
+// GetPolarisBlockByHash retrieves a block from the database by hash, caching it if found.
 func (bc *blockchain) GetPolarisBlockByHash(hash common.Hash) (*types.Block, error) {
```
itsdevbear, line 191: Remove "Polaris"

itsdevbear (commented):
Lgtm, but we need to add hella unit tests.

itsdevbear (approved):
~~I'm good to merge this, but making another linear ticket for the tests. 
~~

Address comments first actually

itsdevbear:
@corduroybera can u make ticket bls

calbera:
So I don’t think we need to merge this one. Implementing it via geth’s txpool provides more features and more testing, we just have to configure it correctly

itsdevbear:
I still think that approach is hood, but if you can find a way to make it seamless I'm on board. 

If anything we should build a wrapper that lives in the eth folder.

calbera:
I was originally doing a wrapper in eth/core/ but I realized there’s no need for it. It works without the wrapper. Either you have the geth TxPool in the blockchain object or you have the plug-in import it directly. The former involves ugly duplication of state

calbera:
https://github.com/berachain/polaris/pull/580/files#diff-732b1d3ebcc7a1ce9c153310d8f40fbd45cbc54b72627b3d298727e6643cf318

PR #572, opened 2023-04-21
feat(erc20): Allow msg.sender to interact with erc20 module

calbera:


itsdevbear:
looks good to me so far, can you add any unit tests that were missing from before as well

itsdevbear (approved):
UtACK, but add more tests

PR #573, opened 2023-04-22
feat(baseapp): Polaris BaseApp

itsdevbear:
We introduce a PolarisBaseApp that chains can import to remove a bunch of the Cosmos-SDK boilerplate bs.


itsdevbear:
to be continued...

PR #574, opened 2023-04-23
refactor(x/evm): Remove the horror show that was offchainkv.

itsdevbear:
image

Note: We should upstream some form of`MountCustomStores` to the sdk, thoughts @tac0turtle


tac0turtle:
yea this makes sense. with the legacy way of writing apps this is easier, depinject is missing lots of features

itsdevbear:
@tac0turtle what's your thoughts on allowing modules to register stores that aren't included as part of the app hash?

PR #583, opened 2023-04-27
feat(txpool): Use Geth TxPool in the TxPool Plugin

calbera:


eth/core/host.go
```diff
@@ -126,15 +126,25 @@ type (
...
-		GetTransaction(common.Hash) *types.Transaction
-		// GetNonce returns the nonce of the given address in the transaction pool.
-		GetNonce(common.Address) (uint64, error)
-		// GetNewTxsEventSubscription returns a subscription with the new txs event channel.
-		GetNewTxsEventSubscription(ch chan<- NewTxsEvent) event.Subscription
+		SendTx(*types.Transaction) error
+		// Pending returns the pending transactions in the transaction pool.
+		Pending(bool) map[common.Address]types.Transactions
+		// Get returns the transaction from the pool with the given hash.
+		Get(common.Hash) *types.Transaction
```
itsdevbear, line 133: No I think what you can do is make it so that all the txpool stuff lives in the eth side and the plugin doesn't need to care,

like basically make it so that on SendTx it does all the mempool stuff or something, then on ProcessTransaction in the blockchain, it does the mempooll removal or soething like that (idk if that exactly works but u catch my drift).

PR #652, opened 2023-05-05
refactor(plugins): Stateless historical plugin

oofncubed:


eth/core/host.go
```diff
@@ -148,20 +150,20 @@ type (
 		// HistoricalPlugin implements `libtypes.Preparable`.
 		libtypes.Preparable
```
calbera, line 151: you can now remove this. basically .Prepare should never be called on the historical plugin

cosmos/x/evm/plugins/historical/historical_data.go
```diff
@@ -38,12 +39,13 @@ import (
 // TODO: WHO WROTE THIS CODE THE FIRST TIME BLS FIX IT IS HORRIBLE.
```
calbera, line 39: you can remove this line bls bls

calbera:
make this historical plugin only. Separate PRs for config, gas, and maybe block

itsdevbear:
Is this good?

oofncubed:
> Is this good?

no, the read paths won't work. the context passed into the Get functions won't be able to unwrap into a sdk.context

itsdevbear:
lets shelve this for now tbh.

PR #654, opened 2023-05-05
refactor(plugin): stateless gas plugin implementation

oofncubed:


itsdevbear:
Is this good?

oofncubed:
> Is this good?

tests pass and mage start still good

itsdevbear:
Can you add a few more tests around consumption and edge cases

itsdevbear:
I think we actually shelf this for now. It feels a little cosmos specific and I don't wanna break stuff kek

PR #662, opened 2023-05-08
feat(log): Support Multiple Coins in SdkConvertCoins

oofncubed:
- Enabling this because there are scenarios where multiple coins need to be captured at once. I.e. sending coins into a LP will require multiple coins being transferred at once, and we cannot assume that transfers will only happen one at at time


itsdevbear:
@calbera confirm this won't brick stuff

itsdevbear:
But otherwise utACK Lgtm

PR #664, opened 2023-05-09
feat(log): Support Multiple Coins in SdkConvertCoins

oofncubed:
redo for: https://github.com/berachain/polaris/pull/662 for proper PR.


cosmos/precompile/bank/helpers.go
```diff
@@ -42,8 +42,8 @@ func sdkCoinsToEvmCoins(sdkCoins sdk.Coins) []generated.IBankModuleCoin {
 	return evmCoins
 }
 
-// extractCoinsFromInput converts coins from input (of type any) into sdk.Coins.
-func extractCoinsFromInput(coins any) (sdk.Coins, error) {
+// ExtractCoinsFromInput converts coins from input (of type any) into sdk.Coins.
```
calbera, line 45: I would move these coins conversion helper functions to the cosmos/lib folder in a file called coins.go or similar

contracts/src/cosmos/precompile/Bank.sol
```diff
@@ -44,27 +44,27 @@ interface IBankModule {
     /**
      * @dev Emitted by the bank module when `amount` tokens are spent by `spender`
      */
-    event CoinSpent(address indexed spender, uint256 amount);
+    event CoinSpent(address indexed spender, Coin[] amount);
```
itsdevbear, line 47: we should make a solidity library that contains Coin so that all the diff precompiles use the same type 

cosmoslib.sol or something like this.

calbera: have to backlog because abigen will recreate the go binding struct with the same name in the same package causing name collision

oofncubed: tracking this via BENG-393
GH issue: https://github.com/berachain/polaris/issues/671

calbera (approved):

PR #689, opened 2023-05-11
feat: Shampoo/beng 108 graphql endpoint

shampoobera:


eth/provider/provider.go
```diff
@@ -21,10 +21,13 @@
 package provider
 
 import (
+	"github.com/ethereum/go-ethereum/cmd/utils"
+	"github.com/ethereum/go-ethereum/eth/ethconfig"
 	"github.com/ethereum/go-ethereum/node"
 
 	"pkg.berachain.dev/polaris/eth/api"
 	"pkg.berachain.dev/polaris/eth/core"
+	"pkg.berachain.dev/polaris/eth/graphql"
```
calbera, line 30: why do we need to include the graphql folder in polaris? can we just import from geth?

shampoobera: ya lemme fix

shampoobera: we can reuse it all from geth

PR #700, opened 2023-05-12
test: solidity struct bytes to go struct

technicallyty:
devbear wanted me to open this PR to show u how i convert solidity struct bytes to Go struct. enjoy.


cosmos/testing/integration/precompile/misc/misc_test.go
```diff
@@ -21,14 +21,73 @@
...
+		To:     big.NewInt(3293),
+		From:   big.NewInt(2385),
+		Amount: big.NewInt(23509),
+	}
+	bzStr := "0x0000000000000000000000000000000000000000000000000000000000000cdd00000000000000000000000000000000000000000000000000000000000009510000000000000000000000000000000000000000000000000000000000005bd5"
+	bz, err := hexutil.Decode(bzStr)
+	Expect(err).To(BeNil())
+
+	// make MsgSendEnergy abi type.
+	msgSendEnergyType, err := abi.NewType("tuple", "", []abi.ArgumentMarshaling{
```
itsdevbear, line 61: @technicallyty do you know if this is possible with the structs generated in `abigen`

itsdevbear: we'd currently have to recreate these everytime, though I guess you could rwite a generator

calbera: can import this type directly from the solidity contract's go bindings, so not needed manually here

technicallyty: @calbera are you referring to the go bindings ala `i_governance_module.abigen.go`? if so, that doesn't work as it's just a Go struct, and abi package doesn't know how to unmarshal abi encoded bytes into Go structs, only abi types. 

unless theres somewhere in abigen thats making these abi types that im not seeing?

calbera: so if you have the .abi.json file of your solidity contract/interface, you can build the abi.ABI struct which has all of this generated for you. Get the Methods --> Inputs/Outputs --> Type

calbera: for the go struct, we have to do the same thing you have here https://github.com/berachain/polaris/blob/main/cosmos/lib/conversions.go#L54-L59

PR #703, opened 2023-05-13
fix(rpc): Backend struct behaviour should match geth more closely. 

itsdevbear:
Trying to more closely match geth, we really need some conformance testing at some point, but for now we mess around.

 @Ferret-san


transmissions12 (approved):
UtAck

PR #704, opened 2023-05-13
feat(testing): OP-stack deployment integration testing.

itsdevbear:


eth/core/state/statedb.go
```diff
@@ -230,10 +230,6 @@ func (sdb *stateDB) GetStorageProof(_ common.Address, _ common.Hash) ([][]byte,
 	return nil, nil
 }
```
calbera, line 231: `GetStorageProof` also needs to be deleted so it calls the Plugin

testing/op-l2.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+SEQ_KEY="73bead3e3e588be3ae77598160d6a5017acdb008973f21166571e400a4b51246"
+L1_RPC="http://localhost:8545"
+RPC_KIND="any"
```
Ferret-san, line 5: `RPC_KIND` should be `basic` to avoid calling non existing methods into the Polaris JSON RPC API

itsdevbear: fixed

eth/core/state/statedb.go
itsdevbear: reeeeee its so cooked LMFAO we need MPT

PR #708, opened 2023-05-15
fix: use operator address for coinbase

oofncubed:


cosmos/x/evm/plugins/block/plugin.go
```diff
@@ -74,7 +77,11 @@ func (p *plugin) GetNewBlockMetadata(number int64) (common.Address, uint64) {
 		panic("block height mismatch")
 	}
 
-	return common.BytesToAddress(cometHeader.ProposerAddress), uint64(cometHeader.Time.UTC().Unix())
+	val, found := p.sk.GetValidatorByConsAddr(p.ctx, cometHeader.ProposerAddress)
+	if !found {
+		panic("validator not found")
```
calbera, line 82: can we include the `cometHeader.ProposerAddress` in the error message

cosmos/x/evm/plugins/block/plugin.go
```diff
@@ -74,7 +77,11 @@ func (p *plugin) GetNewBlockMetadata(number int64) (common.Address, uint64) {
 		panic("block height mismatch")
```
calbera, line 77: here too, more descriptive error msg is more helpful

calbera:
> Fix tests then Lgtm.
> 
> Is there a way we can use not the staking hardtype?

have to get the validator for the operator addr i think

calbera (approved):

PR #709, opened 2023-05-15
chore: move graphql logic to fixture.go

shampoobera:
move http client and function to send GraphQL requests into ```TestFixture``` itself


calbera:
Not using the `EthGraphQLClient` field?

cosmos/testing/integration/fixture.go
```diff
@@ -82,15 +87,18 @@ func NewTestFixture(t network.TestingT) *TestFixture {
 	wsaddr := "ws:" + apiAddr + ":8546"
 	wsClient, _ := ethclient.DialContext(ctx, wsaddr)
 
+	graphQLClient := &http.Client{}
```
calbera, line 90: unused

cosmos/testing/integration/fixture.go
```diff
@@ -132,6 +140,38 @@ func (tf *TestFixture) CreateKeyWithName(name string) {
...
+
+	client := &http.Client{}
+
+	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(requestBody))
+	if err != nil {
+		return "", 500, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := client.Do(req)
```
calbera, line 160: rather than rewriting http connection logic, try using the Geth's built-in rpc client 
https://github.com/berachain/polaris-geth/blob/stateful-v1.11.4/rpc/client.go#L166

shampoobera: GraphQL is an alternative to RPC, won't work i believe

https://eips.ethereum.org/EIPS/eip-1767

itsdevbear (approved):

PR #721, opened 2023-05-16
feat(statedb): Per block logs journal

calbera:
- Allow the logs journal to store the entire block's logs, similar to Geth


eth/core/processor.go
```diff
@@ -242,20 +235,16 @@ func (sp *StateProcessor) Finalize(
...
-	var logs []*types.Log
-	for _, receipt := range sp.receipts {
-		receipt.BlockHash = blockHash
-		for _, log := range receipt.Logs {
-			log.BlockHash = blockHash
-		}
-		logs = append(logs, receipt.Logs...)
+
+	// Set the missing fields on receipts effective gas price and block hash.
+	if err := sp.receipts.DeriveFields(
```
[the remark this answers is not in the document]
calbera: but we iterate through the receipts -> to logs either way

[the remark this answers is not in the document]
itsdevbear: I would mirror geth.

eth/core/state/interfaces.go
```diff
@@ -101,6 +101,8 @@ type (
 		Logs() []*coretypes.Log
 		// GetLogs returns the logs of the tx with the given metadata.
 		GetLogs(hash common.Hash, blockNumber uint64, blockHash common.Hash) []*coretypes.Log
+		// GetBlockLogsAndClear returns the logs of the block and clears the journal.
+		GetBlockLogsAndClear(blockHash common.Hash) []*coretypes.Log
```
itsdevbear, line 105: we need to stop adding shit not in geth keke

calbera: better separation of concerns to have the block logs list in the logs journal imo, especially based on our abstractions. geth stores logs per block as well

itsdevbear: why do we need an Clear then?

just clear on StateDB Reset

itsdevbear: That's how Geth does it tho lmeow

calbera: lmao wow geth never clears the logs map

calbera: we can also not clear but I think its a good optimization

itsdevbear:
Mirror geth good project for now. Optmize later for sure

calbera:
working off of #734, should be merged only after that PR is in

calbera:
Think its worth shelving this for now. Storing logs per block is definitely more inefficient than what we have currently and does not add any new functionality. This change is more of a refactor that we can avoid for now.

will address the receipts storage issue in another PR.

PR #729, opened 2023-05-17
fix(core): fix historical plugin? optimize receipts?

calbera:
[WIP] chain does not work without tx, block, receipts caches in the blockchain object. Need to correct the historical plugin logic so that chain works even if cache not supported.


itsdevbear:
Honestly a good time to add the debug mode

PR #732, opened 2023-05-18
feat(testing): Arch/beng 449 scope out usage of hive ethereum

archbear:


cosmos/x/evm/keeper/genesis.go
```diff
@@ -41,10 +43,14 @@ func (k *Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) error
 		}
 	}
 
-	// Start the polaris "Node" in order to spin up things like the JSON-RPC server.
-	if err := k.polaris.StartServices(); err != nil {
-		return err
-	}
+	go func() {
+		time.Sleep(2 * time.Second) //nolint:gomnd // we will fix this eventually.
```
itsdevbear, line 47: @calbera the reason we do this is because hive will automatically start spamming the node as soon as it detects that 8545 is up, so we actually want to add a short delay until the chain is ready.

This is ghetto, but it does work, can you think of a better way?

calbera: not really

calbera: just need to make sure if the Node start fails, returns an error, we panic and halt chain

itsdevbear: I'm thinking we add a `rpc delay` to `polaris.toml` thoughts?

PR #734, opened 2023-05-18
fix(statedb): Correctly implement StateDB's deep copy

calbera:
- Correct statedb behavior in accordance to geth 
- Simplify Polaris interfaces where possible


calbera:
getting the build error: "import cycle not allowed in test". any idea why?
```
ERROR Running error: context loading failed: failed to load packages: 
failed to load with go/packages: internal error: go list gives conflicting information for package 
pkg.berachain.dev/polaris/eth/core/state/journal [pkg.berachain.dev/polaris/eth/core/state/journal.test]
```

eth/core/processor.go
```diff
@@ -239,6 +239,7 @@ func (sp *StateProcessor) BuildAndRegisterPrecompiles(precompiles []precompile.R
 		if err != nil {
 			panic(err)
 		}
+		// TODO: set code on the statedb for every precompiled contract.
```
itsdevbear, line 242: nice

eth/core/state/interfaces.go
```diff
@@ -38,6 +37,8 @@ type Plugin interface {
 	libtypes.Preparable
 	// Reset resets the state with the given `context`.
 	libtypes.Resettable
+	// Plugin implements `libtypes.Cloneable`.
```
itsdevbear, line 40: cloneable good proj

eth/core/state/journal/logs.go
```diff
@@ -25,21 +25,38 @@ import (
 	coretypes "pkg.berachain.dev/polaris/eth/core/types"
 	"pkg.berachain.dev/polaris/lib/ds"
 	"pkg.berachain.dev/polaris/lib/ds/stack"
+	libtypes "pkg.berachain.dev/polaris/lib/types"
+	"pkg.berachain.dev/polaris/lib/utils"
 )
 
+// LogsI defines the interface for tracking logs created during a state transition.
+type LogsI interface {
```
itsdevbear, line 33: using the `I` terminology is frowned upon. StateDBLogs?

eth/core/state/journal/refund.go
```diff
@@ -79,3 +92,15 @@ func (r *refund) RevertToSnapshot(id int) {
 func (r *refund) Finalize() {
 	r.Stack = stack.New[uint64](initCapacity)
 }
+
+// Clone implements `libtypes.Cloneable`.
+func (r *refund) Clone() RefundI {
```
itsdevbear, line 97: `I` notatation everywhere

eth/core/state/journal/suicides.go
```diff
@@ -43,6 +45,19 @@ type suicideStatePlugin interface {
 	SubBalance(common.Address, *big.Int)
 }
 
+type SuicidesI interface {
```
itsdevbear, line 48: +1

itsdevbear (commented):
Overall makes sense to me, though you removed a bunch of tests lol.

I think we should just make the Controller Cloneable, thoughts?

cosmos/x/evm/plugins/precompile/plugin.go
```diff
@@ -164,7 +163,14 @@ func (p *plugin) Run(
 // EnableReentrancy sets the state so that execution can enter the EVM again.
 //
 // EnableReentrancy implements core.PrecompilePlugin.
-func (p *plugin) EnableReentrancy(_ context.Context) {
+func (p *plugin) EnableReentrancy(evm ethprecompile.EVM) {
+	sdb := utils.MustGetAs[vm.PolarisStateDB](evm.GetStateDB())
```
itsdevbear, line 167: Can we move the re-entrancy guard to the `eth` folder?

calbera: this is logic specific to Cosmos

calbera: that we need to stop/start emitting logs, switch gas config

itsdevbear: Check slack my bad

calbera:
didnt remove tests, and yeah I was thinking Controller could be cloneable but massive lift, too much for now

itsdevbear:
I guess controller you end up in generic hell

calbera:
I can add a couple more unit tests though

itsdevbear:
Add some tests that ensure that if you modify the clones it doesn't modify the original 🫡

calbera:
should be good to go, the pending nonce integration test thats failing is unaffected by this change, which btw it passes for me locally

calbera:
Just verified in geth that `.Copy()` is only called on the StateDB for read-only things like traces and is not committed.
image

itsdevbear (approved):

itsdevbear:
we can cleanup later

PR #740, opened 2023-05-23
feat(cosmos): Mint Module Precompile

itsdevbear:


itsdevbear:
@corduroybera this is low priority but can you get someone to write tests for these.

itsdevbear:
close for now, we don't need more things to worry about.

PR #743, opened 2023-05-23
feat(testing): Docker testing framework

itsdevbear:


itsdevbear:
make sure we implement `Error()` correctly as well, we will probably want to have errors at both the plugin and db level so that will need to be properly managed as well.

PR #744, opened 2023-05-24
feat(flashbera): [WIP] 4 shampoo

itsdevbear:


calbera:
I think the flashbera APIs, `BundleAPI`, `PrivateTxBundleAPI` belong in polaris-geth.
The implementations of the backend functions belong in Polaris imo

itsdevbear:
i'm not losing sleep either way.

calbera:
CallBundle and BundleAPI already exists in polaris-geth

PR #749, opened 2023-05-25
chore(cosmos): Upgrade to CometBFT 0.38

itsdevbear:
TODO:
- [ ] Get rid of `from` field in EthTransactionRequest, 0 chance I let that bech32 bs in here.


cosmos/proto/polaris/evm/v1alpha1/tx.proto
```diff
@@ -44,9 +44,13 @@ service MsgService {
 
 // EthTransactionRequest encapsulates an Ethereum transaction as an SDK message.
 message EthTransactionRequest {
+  option (cosmos.msg.v1.signer) = "from";
   option (gogoproto.goproto_getters) = false;
   // data is inner transaction data of the Ethereum transaction
   bytes data = 1;
+
+  // from is the sender of the transaction
+  string from = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
```
itsdevbear, line 53: @tac0turtle is there a way to avoid this? I hate that it's forcing me to use bech32 with the eth transaction

cosmos/x/evm/keeper/abci.go
```diff
@@ -26,18 +26,14 @@ import (
...
 
-// Precommit is called during the Commit processing of the ABCI lifecycle, right before the state
-// is committed to the root multistore.
-func (k *Keeper) Precommit(ctx context.Context) {
+func (k *Keeper) EndBlock(ctx context.Context) error {
 	// Finalize the Polaris Ethereum block.
-	if err := k.polaris.Finalize(ctx); err != nil {
-		panic(err)
-	}
+	return k.polaris.Finalize(ctx)
```
calbera, line 38: now being called in EndBlock?

itsdevbear: yeah cause comet lifecycle changed.

PR #750, opened 2023-05-25
refactor(core): adhere blockhashes to match geth

calbera:
moving towards matching the [miner/worker](https://github.com/berachain/polaris-geth/blob/stateful-v1.11.4/miner/worker.go) model in geth.


eth/core/processor.go
```diff
@@ -139,20 +141,17 @@ func (sp *StateProcessor) Prepare(evm *vm.GethEVM, header *types.Header) {
...
 
 	// Set the transaction context in the state database.
 	// This clears the logs and sets the transaction info.
 	sp.statedb.SetTxContext(tx.Hash(), len(sp.txs))
 
 	// Inshallah we will be able to apply the transaction.
 	receipt, result, err := ApplyTransactionWithEVMWithResult(
-		sp.evm, sp.cp.ChainConfig(), nil, &gasPool, sp.statedb, sp.header, tx, &gasUsed,
+		sp.evm, sp.cp.ChainConfig(), gasPool, sp.statedb, sp.header.BaseFee,
+		sp.header.Number, sp.sealhash, tx, &sp.header.GasUsed,
```
calbera, line 154: way more efficient to precompute the block hash (doing the full rlpHash algo) once per block rather than for every tx

eth/core/processor.go
```diff
@@ -68,6 +68,7 @@ type StateProcessor struct {
 	// during the processing of transactions. This allows us to utilize this information to
 	// build the `block` and return the canonical receipts in `Finalize`.
 	header   *types.Header
+	sealhash common.Hash // hash of the block prior to being sealed (Finalized)
```
calbera, line 71: geth terminology

itsdevbear (approved):
yeah yolo im good with it

eth/core/processor.go
itsdevbear: I wouldn't recommend having the state machine modify the unit64 on the header, that is some hood shit.

itsdevbear (changes requested):
don't pass the header gasUSed into the state machine thats hood af

eth/core/processor.go
calbera: thats how geth does it lol

calbera:
> don't pass the header gasUSed into the state machine thats hood af

https://github.com/berachain/polaris-geth/blob/stateful-v1.11.4/miner/worker.go#L868

eth/core/processor.go
itsdevbear: Rly lol link??

calbera: https://github.com/berachain/polaris-geth/blob/stateful-v1.11.4/miner/worker.go#L868

itsdevbear: So hood. 

Do we wanna just leave the state processor as is and start building a Miner struct on the side?

calbera: also https://github.com/flashbots/builder/blob/main/internal/ethapi/api.go#L2309

calbera: its weird but thats why it takes the pointer to integer

calbera: I think we can leave this as is and merge if we want to. the miner/worker stuff would require way more upheaval so again smth thats prob best for later

itsdevbear:
replaced with https://github.com/berachain/polaris/pull/764

PR #767, opened 2023-05-30
refactor(evm): Deprecate `x/bank`

itsdevbear:


cosmos/x/evm/plugins/state/keys.go
```diff
@@ -27,6 +27,13 @@ import (
 
 // NOTE: we use copy to build keys for max performance: https://github.com/golang/go/issues/55905
 
+func BalanceKeyFor(address common.Address) []byte {
+	bz := make([]byte, 1+common.AddressLength)
+	copy(bz, []byte{types.BalanceKeyPrefix})
+	copy(bz[1:], address[:])
+	return bz
```
calbera, line 34:
```suggestion
	return append([]byte{types.BalanceKeyPrefix}, address[:]...)
```
the `[:]` also does a copy

cosmos/x/evm/genesis.go
```diff
@@ -22,45 +22,53 @@ package evm
...
 // ValidateGenesis performs genesis state validation for the evm module.
-func (AppModuleBasic) ValidateGenesis(cdc codec.JSONCodec, _ client.TxEncodingConfig, bz json.RawMessage) error {
-	var data types.GenesisState
-	if err := cdc.UnmarshalJSON(bz, &data); err != nil {
-		return fmt.Errorf("failed to unmarshal %s genesis state: %w", types.ModuleName, err)
-	}
-	return types.ValidateGenesis(data)
+func (AppModuleBasic) ValidateGenesis(_ codec.JSONCodec, _ client.TxEncodingConfig, bz json.RawMessage) error {
+	ethGen := new(core.Genesis)
+	return ethGen.UnmarshalJSON(bz) // todo: improve
```
calbera, line 48: doesnt validate

calbera:
I do think there should be a "bank store" abstraction, which implements our expected bank keeper interface everywhere. rn we only have one denom so just the `balances: addr -> big.Int` is fine

cosmos/precompile/auth/auth_test.go
```diff
@@ -283,12 +283,12 @@ var _ = Describe("Address Precompile", func() {
...
 				evm,
 				common.Address{},
-				big.NewInt(0),
+				new(big.Int),
 				false,
 				granter,
 				grantee,
 				sdkCoinsToEvmCoins(limit),
-				big.NewInt(0),
+				new(big.Int),
```
calbera, line 291: we should do separate changes, like this refactor, in diff PRs

cosmos/x/evm/plugins/state/keys.go
```diff
@@ -75,12 +82,12 @@ func CodeKeyFor(codeHash common.Hash) []byte {
...
-func CodeHashFromKey(key []byte) common.Hash {
-	return common.BytesToHash(key[1:])
-}
-
 // AddressFromCodeHashKey returns the address from a code hash key.
 func AddressFromCodeHashKey(key []byte) common.Address {
 	return common.BytesToAddress(key[1:])
 }
+
+// AddressFromBalanceKey returns the address from a code hash key.
```
calbera, line 90:
```suggestion
// AddressFromBalanceKey returns the address from a balance key.
```

cosmos/precompile/auth/auth_test.go
itsdevbear: for future yeah, but yolo for now.

itsdevbear:
good otherwise @calbera we will
 block genesis stuff in a second pr

cosmos/x/evm/plugins/state/genesis.go
```diff
@@ -21,70 +21,65 @@
...
+		// TODO: technically wrong since its overriding / hacking the auth keeper and
+		// we are using the nonce from the account keeper as well.
+		p.CreateAccount(address)
+		p.SetBalance(address, account.Balance)
+		if account.Code != nil {
+			p.SetCode(address, account.Code)
+			for k, v := range account.Storage {
+				p.SetState(address, k, v)
+			}
 		}
```
calbera, line 48:
```suggestion
		if account.Code != nil {
			p.SetCode(address, account.Code)
		}
		if account.Storage != nil {
			for k, v := range account.Storage {
				p.SetState(address, k, v)
			}
		}
```

calbera (approved):

calbera (approved):

PR #769, opened 2023-05-31
fix(rpc): Resolve height in the future issue patch.

itsdevbear:
Known regression when you query with height 0. It was actually returning incorrectly before, so I'd rather it error than return incorrectly. This can be fixed after the bank removal is complete.


sergical (approved):
Verified `delegate` works with MM with no errors

PR #770, opened 2023-05-31
fix(txpool): Return correct pending nonce for both cosmos and eth txs

calbera:
fixing issue:
"the case in which it could return the wrong pending nonce is if you send in E, C, E txs. Say the first Eth tx sends in a nonce x. The next tx, Cosmos tx has a sequence/nonce of x+1. Now before sending the next tx, the eth client asks for pending nonce and currently, it will still return x+1 as pending because its only using the eth txs. So now you have 2 txs with nonce x+1 , so depending on ordering one of them will fail during processing."


oofncubed:
Change panic message to b more descriptive

oofncubed (approved):

itsdevbear (approved):

PR #771, opened 2023-05-31
eip1155 chain-id updated from 69420 -> 2061

garrettparris:
Keplr parses evm chain Id based on the ending numbers in the cosmos chain-id. As a temp solution prior to submitting a refactoring PR to keplr wallet in order to change this, the last four digits of our cosmos chain id now matches our evm chain id. This change is required for keplr to submit eth transactions to our JSON-RPC endpoint.


oofncubed (approved):
utACK

PR #775, opened 2023-05-31
feat(txpool): Use Geth TxPool for the Cosmos App-side mempool

calbera:


cosmos/x/evm/plugins/txpool/mempool/mempool.go
```diff
@@ -22,74 +22,55 @@ package mempool
...
+	gtp.cp = cp
+	gtp.serializer = serializer
 }
 
-// SetBaseFee updates the base fee in the priority policy.
-func (etp *EthTxPool) SetBaseFee(baseFee *big.Int) {
-	etp.priorityPolicy.baseFee = baseFee
+// Prepare updates the mempool for the current block. Set the block number, block time, and base
+// fee.
+func (gtp *WrappedGethTxPool) Prepare(blockNumber *big.Int, blockTime uint64, baseFee *big.Int) {
```
itsdevbear, line 72: why not just hook this up to the NewChainHead channel?

calbera: that gets emitted after the end of the prev block, in which case the current block data has already updated.

eth/core/chain_writer.go
```diff
@@ -92,8 +92,8 @@ func (bc *blockchain) Prepare(ctx context.Context, number uint64) {
 		Nonce:      types.BlockNonce{},
 	}
 
-	// We update the base fee in the txpool to the next base fee.
-	bc.tp.SetBaseFee(header.BaseFee)
+	// We update the txpool with the new block information.
+	bc.tp.Prepare(header.Number, header.Time, header.BaseFee)
```
itsdevbear, line 96: tbh I think we can remove all this Prepare() stuff and use channels, I lowkey think more efficient too.

itsdevbear:
@calbera i would recommend redoing this off of my `deprecate-bank-keeper` branch

this is gonna be a nightmare to resolve merge conflicts if you don't

calbera:
- handler to live in the evm keeper, start/stop/broadcast condition
- Set client context on handler, not plugin
- Fix select to iterate local then remote (parallelize?)

itsdevbear:
I think handler maybe on the baseapp?

eth/core/chain_reader.go
```diff
@@ -218,6 +218,11 @@ func (bc *blockchain) GetBlockByHash(hash common.Hash) *types.Block {
 
 // GetBlock retrieves a block from the database by hash and number, caching it if found.
 func (bc *blockchain) GetBlockByNumber(number uint64) *types.Block {
+	// return the genesis block if querying block 0
+	if number == 0 {
+		return types.NewBlockWithHeader(bc.genesisBlock.Header())
```
itsdevbear, line 223: return the stored genesis block on the object

cosmos/x/evm/plugins/txpool/handler/handler.go
```diff
@@ -0,0 +1,125 @@
...
+			// TODO: log error?
+			panic(errorslib.Wrap(err, "failed to serialize transaction"))
+		}
+
+		// Send the transaction to the CometBFT mempool, which will gossip it to peers via
+		// CometBFT's p2p layer.
+		syncCtx := h.clientCtx.WithBroadcastMode(flags.BroadcastSync)
+		rsp, err := syncCtx.BroadcastTx(txBytes)
+		if rsp != nil && rsp.Code != 0 {
+			// TODO: log error?
```
itsdevbear, line 117: i think for broadcast fail, we could throw into a retry loop

PR #776, opened 2023-05-31
fix(staking): Return correctly formatted structs for UnbondingDelegationEntry, RedelegationEntry

calbera:
fix for 
image


itsdevbear (approved):
god our types are so fucked, we need precompile modules

oofncubed (approved):

PR #783, opened 2023-06-03
feat(precompile): Add re-entrancy lock to container.

itsdevbear:
TODO: make sure that there isn't cross talk between `eth_calls` and actually block execution.


itsdevbear (commented):
to ensure no cross talk, just out the lock on the context ez pz

itsdevbear:
probably don't need and this impl was broken anyways

PR #789, opened 2023-06-06
fix(rpc): temp rpc fix

itsdevbear:


itsdevbear:
@BrickBera we may need to get @archbear's PR in to totally fix this, but this might help.

itsdevbear:
@calbera fresh branch, this one is broken

PR #790, opened 2023-06-06
Debugging App Hash

calbera:


itsdevbear:
id add more logs like absolutely everywhere

PR #787, opened 2023-06-05
feat: implement x/evm/block genesis

archbear:

PR #794, opened 2023-06-07
feat(evm): Read and store Genesis block

archbear:
deprecating https://github.com/berachain/polaris/pull/787 since the pr is broken


calbera (commented):
set genesis block/load state from disk in eth/core in next PR

calbera (approved):

itsdevbear (approved):

PR #806, opened 2023-06-09
feat(evm): runtime module setup

itsdevbear:


itsdevbear:
TODO: 
- [ ] Get rid of ugly race condition sleeps
- [ ] Remove circular evm keeper reference required for the polar.Polaris struct (we only need this for testing, polar.Polaris should be fully removed from the EVMKeeper.
- [ ] Rename EVMKeeper -> PolarisKeeper

calbera (approved):

PR #810, opened 2023-06-11
feat(cosmos): `runtime` module + `WrappedGethTxPool`

itsdevbear:


itsdevbear:
todo: 
- build header in prepare/process proposal, probably will have to be done in both (in theory the header should be gossiped but whatever). 
this will remove Prepare() from the core.Blockchain

itsdevbear:
@tac0turtle is there a way to have comet gossip some arbitrary data, so that we can have the proposer send the ethereum block header along with its proposal?

tac0turtle:
there is not right now unless you fork comet. The change wouldnt be hard. Abci is limiting for anything p2p related

itsdevbear:
replace by #990

PR #814, opened 2023-06-12
feat(BlockPlugin): block plugin support of GetHeaderByHash

archbear:
This PR fixes Block/Historical Plugin code paths so that:
- all queries for header go through block plugin
- all queries for block go through historical plugin


itsdevbear:
You don't wanna store anything by hash in the IAVL tree since it will cause state growth.

itsdevbear:
If we really wanted we could do a 256 block revolving queue, but idk if thats the move. Whats the issue that this PR is trying to solve?

archbear:
> If we really wanted we could do a 256 block revolving queue, but idk if thats the move. Whats the issue that this PR is trying to solve?

The get genesis by header and by number rpc queries were causing the backend to call the historical plugin instead of the block plugin. That lead to trying to distinguish the responsibilities of the block and historical plugin from header to txs etc.

eth/polar/backend.go
```diff
@@ -218,21 +218,22 @@ func (b *backend) HeaderByNumberOrHash(ctx context.Context,
...
-	return b.polar.blockchain.GetHeaderByHash(hash), nil
+	header := b.polar.blockchain.GetHeaderByHash(hash)
+	if header == nil {
+		return nil, errors.New("header for hash not found")
+	}
+
+	// if blockNrOrHash.RequireCanonical && b.eth.blockchain.GetCanonicalHash(header.Number.Uint64()) != hash {
+	// 	return nil, errors.New("hash is not currently canonical")
+	// }
+	return header, nil
```
calbera, line 236:
```suggestion
	return b.polar.blockchain.GetHeaderByHash(hash)
```

eth/core/chain_reader.go
```diff
@@ -116,11 +116,11 @@ func (bc *blockchain) CurrentSafeBlock() *types.Header {
...
 // found.
 func (bc *blockchain) GetHeaderByHash(hash common.Hash) *types.Header {
-	block := bc.GetBlockByHash(hash)
-	if block == nil {
+	header, err := bc.bp.GetHeaderByHash(hash)
+	if header == nil || err != nil {
 		return nil
 	}
-	return block.Header()
+	return header
```
calbera, line 123: `return bc.bp.GetHeaderByHash(hash)`

cosmos/x/evm/types/keys.go
```diff
@@ -29,8 +29,10 @@ const (
 	BlockHashKeyToReceiptsPrefix
 	TxHashKeyToTxPrefix
 	VersionKey
-	HeaderKey
-	GenesisHeaderKey
+	HeaderNumKey
+	HeaderHashKey
```
calbera, line 33:
```suggestion
	HeaderHashKeyPrefix
```

calbera (approved):
tACK

calbera:
@itsdevbear this is good to go, integration tests are very flaky rn. also broken on main

eth/core/chain_reader.go
```diff
@@ -38,14 +38,14 @@ type ChainReader interface {
 // ChainBlockReader defines methods that are used to read information about blocks in the chain.
 type ChainBlockReader interface {
 	CurrentHeader() *types.Header
-	CurrentBlock() *types.Header
```
itsdevbear, line 41: this does not match geth.

eth/polar/backend.go
```diff
@@ -398,7 +393,7 @@ func (b *backend) GetLogs(
 // GetTd returns the total difficulty of a block in the canonical chain.
 // This is hardcoded to 69, as it is only applicable in a PoW chain.
 func (b *backend) GetTd(_ context.Context, hash common.Hash) *big.Int {
-	if header := b.polar.blockchain.GetHeaderByHash(hash); header != nil {
+	if header, err := b.polar.blockchain.GetHeaderByHash(hash); err == nil {
```
itsdevbear, line 396: doesn't match geth

eth/core/chain_reader.go
calbera: Same function

eth/polar/backend.go
calbera: Just propogating error, useful for the client making call

calbera:
@itsdevbear going to close this branch and the block plugin refactor should be on @archbear’s new branch. But we can revert some of the new variances with geth. Was mainly for passing the error up the stack to the RPC call

cosmos/x/evm/plugins/block/header_test.go
```diff
@@ -39,70 +46,108 @@ var _ = Describe("Header", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		Expect(header2.Hash()).To(Equal(header.Hash()))
+
+		// get unknown header should return the latest header (10)
+		header3, err := p.GetHeaderByNumber(11)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(header3.Hash()).To(Equal(header.Hash()))
+	})
+
+	FIt("should be able to prune headers", func() {
```
itsdevbear, line 97: focus

itsdevbear (commented):
add test for pruning

itsdevbear (approved):

itsdevbear:
@calbera why removing hive from ci?

calbera:
@itsdevbear cuz its just stuck like this 
image
re-enable in the other PR

itsdevbear (approved):

PR #837, opened 2023-06-20
feat(precompile): Enforce read-only calls from the EVM cannot modify state

calbera:


eth/core/state/journal/suicides.go
```diff
@@ -83,10 +87,19 @@ func (s *suicides) RegistryKey() string {
...
+// SetReadOnly sets the read-only mode of the journal.
+func (s *suicides) SetReadOnly(readOnly bool) {
+	s.readOnly = readOnly
+}
+
 // Suicide implements the PolarisStateDB interface by marking the given address as suicided.
 // This clears the account balance, but the code and state of the address remains available
 // until after Commit is called.
 func (s *suicides) Suicide(addr common.Address) bool {
+	if s.readOnly {
```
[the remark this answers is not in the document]
calbera: image
true the evm already checks for it

cosmos/x/evm/plugins/precompile/plugin.go
```diff
@@ -119,30 +120,54 @@ func (p *plugin) SetTransientKVGasConfig(transientKVGasConfig storetypes.GasConf
...
+			// NOTE: this only propagates an error back to the EVM if the type of the given panic
+			// value is error or string (any other type of panic value is ignored)
+			var isError bool
+			if err, isError = utils.GetAs[error](panicked); !isError {
+				if strErr, isString := utils.GetAs[string](panicked); isString {
+					err = errors.New(strErr)
+				}
+			}
+		}
+	}()
```
calbera, line 158: Also allows any arbitrary panic in precompiles to be caught here so that
- method handler does not crash
- tx gets included, but will have a vm error (also more descriptive for devs)
- gas actually gets consumed for a bad precompile call (causing a panic)

cosmos/x/evm/plugins/precompile/plugin.go
```diff
@@ -119,41 +120,69 @@ func (p *plugin) SetTransientKVGasConfig(transientKVGasConfig storetypes.GasConf
...
 	p.enableReentrancy(sdb)
 
 	// handle overconsumption of gas
 	if gm.GasConsumed() > suppliedGas {
-		return nil, 0, vm.ErrOutOfGas
+		suppliedGas = gm.GasConsumed() // ensures 0 gas remaining if gas was overconsumed
+		if err == nil {
+			// only return out of gas error if no other error occurred
+			err = vm.ErrOutOfGas
+		}
```
calbera, line 182: should out of gas error take priority over all other errors? imo its more accurate to return the other errors (like ExecutionReverted, WriteProtection, etc.) bc the evm specifically handles these

itsdevbear: what does geth do?

calbera: geth consumes gas per opcode, so if it hits out of gas after any opcode, it will just return it. we wait till the end of the execution to check for overconsumption of gas

cosmos/x/evm/store/snapmulti/store.go
```diff
@@ -38,63 +39,78 @@ const (
...
-	// get kvstore from mapMultiStore and set cachekv to memory
-	cms[key] = cachekv.NewStore(s.GetCommittedKVStore(key))
 	return cms[key]
 }
 
 // Snapshot implements `libtypes.Snapshottable`.
-func (s *store) Snapshot() int {
-	cms := s.journal.Peek()
-	if cms == nil {
+func (s *Store) Snapshot() int {
```
itsdevbear, line 111: why public `Store` now?

cosmos/x/evm/plugins/precompile/plugin.go
```diff
@@ -119,42 +120,62 @@ func (p *plugin) SetTransientKVGasConfig(transientKVGasConfig storetypes.GasConf
...
-	caller common.Address, value *big.Int, suppliedGas uint64, readonly bool,
-) ([]byte, uint64, error) {
-	// use a precompile-specific gas meter for dynamic consumption
-	gm := storetypes.NewInfiniteGasMeter()
-	// consume static gas from RequiredGas
-	gm.ConsumeGas(pc.RequiredGas(input), "RequiredGas")
-
-	// get native Cosmos SDK context from the Polaris StateDB
+	caller common.Address, value *big.Int, suppliedGas uint64, readOnly bool,
+) (ret []byte, gasRemaining uint64, err error) {
```
itsdevbear, line 124: named errors bad project imo, find a way to not have to do this.

.golangci.yaml
```diff
@@ -264,7 +264,7 @@ linters:
     - nilnil # checks that there is no simultaneous return of nil error and an invalid value
     - noctx # finds sending http request without context.Context
     - nolintlint # reports ill-formed or insufficient nolint directives
-    - nonamedreturns # reports all named returns
+    # - nonamedreturns # reports all named returns
```
itsdevbear, line 267: make this linter pass please

itsdevbear (changes requested):
a few things

cosmos/x/evm/plugins/precompile/plugin.go
calbera: don't think its _not_ gothonic actually. cosmos-sdk and geth both do this for panic recovery. its really the only way

cosmos/x/evm/store/snapmulti/store.go
calbera: true

cosmos/x/evm/plugins/precompile/plugin.go
itsdevbear: okay re-enable the linter though and just nolint it here for now

.golangci.yaml
```diff
@@ -186,7 +186,7 @@ linters-settings:
   nakedret:
     # Make an issue if func has more lines of code than this setting, and it has naked returns.
     # Default: 30
-    max-func-lines: 0
```
itsdevbear, line 189: 100 lines is kinda long. What do you need this for?

cosmos/x/evm/plugins/precompile/plugin.go
```diff
@@ -117,44 +118,66 @@ func (p *plugin) SetTransientKVGasConfig(transientKVGasConfig storetypes.GasConf
...
+		defer func() { ms.SetReadOnly(false) }()
+	}
 
-	// disable reentrancy into the EVM
+	// disable reentrancy into the EVM only during precompile execution
 	p.disableReentrancy(sdb)
+	defer p.enableReentrancy(sdb)
+
+	// recover from any panic during precompile execution for the EVM to handle as a vm error
+	defer func() {
```
itsdevbear, line 144: I think we should abstract this into a panic handler, and then pass pointers in

cosmos/x/evm/store/cachekv/readonly_store.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	storetypes "cosmossdk.io/store/types"
+
+	"pkg.berachain.dev/polaris/eth/core/vm"
+)
+
+var _ storetypes.CacheKVStore = (*ReadOnlyStore)(nil)
+
+// ReadOnlyStore is a wrapper around cachekv.Store that panics on any write operation.
+type ReadOnlyStore struct {
```
itsdevbear, line 32: wait , remind me why we can't use a saved err here?

cosmos/x/evm/store/snapmulti/store.go
```diff
@@ -38,7 +39,7 @@ const (
 // corresponding cache kv store currently being used.
 type mapMultiStore map[storetypes.StoreKey]storetypes.CacheKVStore
 
-// store is a wrapper around the Cosmos SDK `MultiStore` which supports snapshots and reverts.
+// Store is a wrapper around the Cosmos SDK `MultiStore` which supports snapshots and reverts.
```
itsdevbear, line 42:
```suggestion
// store is a wrapper around the Cosmos SDK `MultiStore` which supports snapshots and reverts.
```

eth/core/precompile/errors.go
```diff
@@ -49,4 +49,7 @@ var (
 	// ErrNoPrecompileMethodForABIMethod is returned when no precompile method is provided for a
 	// corresponding ABI method.
 	ErrNoPrecompileMethodForABIMethod = errors.New("this ABI method does not have a corresponding precompile method")
+
+	// ErrPanic is returned when a precompile method panics.
+	ErrPanic = errors.New("panic cought during execution of precompile")
```
itsdevbear, line 54: `ErrPrecompilePanic`

eth/core/precompile/stateful.go
```diff
@@ -89,16 +89,26 @@ func (sc *stateful) Run(
...
 		ctx,
 		evm,
 		caller,
 		value,
-		readonly,
 		unpackedArgs...,
 	)
 
+	// Ensure that no logs were emitted during the execution of the precompile if in read-only mode.
+	if readonly && len(evm.GetStateDB().Logs()) > logsLen {
```
itsdevbear, line 108: I think there is a nicer way to handle this using a TryLock or Sempahore or something, opposed to using log length.

.golangci.yaml
calbera: For the precompile plugin’s Run function which is way more than the default of 30 lines. 100 is arbitrary though

cosmos/x/evm/plugins/precompile/plugin.go
calbera: Fair

cosmos/x/evm/store/cachekv/readonly_store.go
calbera: Geth catches the error as soon as a write op code is called. Similarly here we can catch it as soon as a write happens. The panic handler also catches other precompile panics as well

eth/core/precompile/stateful.go
calbera: Sure

calbera:
image
@itsdevbear EVM halts execution on these errors so waiting till the end to check for over consumption of gas is not consistent

itsdevbear:
Yeah makes sense @calbera I still think we wanna nize the panic handler if we can, there has to be a clean way.

calbera:
I think the choice is to be clean (finish execution and check for the errors) or consistent with EVM (stop execution immediately and return error).  

- [x] `ErrExecutionRevert`: we return it as soon as we catch an error so its good
- [ ] `ErrOutOfGas`: we wait till the end to check. can use the limited gas meter and panic to halt early
- [ ] `ErrWriteProtection`: we can panic on writes/add logs to return immediately or set flags to check at the end

cosmos/lib/evm.go
```diff
@@ -42,6 +48,10 @@ func DeployOnEVMFromPrecompile(
 	contractCode string, // hex-encoded string
 	constructorArgs ...any,
 ) (common.Address, []byte, error) {
+	if utils.MustGetAs[precompile.MultiStore](ctx.MultiStore()).IsReadOnly() {
+		return common.Address{}, nil, vm.ErrWriteProtection
```
calbera, line 52: realized bc we return an error here, this will actually get wrapped with `ErrExecutionReverted` which could be wrong

cosmos/lib/evm.go
```diff
@@ -72,6 +82,10 @@ func CallEVMFromPrecompile(
 	methodName string,
 	args ...any,
 ) ([]byte, error) {
+	if utils.MustGetAs[precompile.MultiStore](ctx.MultiStore()).IsReadOnly() {
+		return nil, vm.ErrWriteProtection
```
calbera, line 86: +1

itsdevbear:
I think we can move away from KVstore in our own statedb, and use something that propagates the error opposed to panicking, this can simply be a wrapper around kvstore imo. fuck collections tbh.

itsdevbear:
for ErrWriteProtection same thing right? the opcode would error versus we don't have that luxury? 

Maybe we create an evm operations manager that is a middleware between the keeper/context that happens all this shit, and it's handling "opcodes"

itsdevbear:
the concept of a precompile "interpreter" isn't the worst

cosmos/lib/evm.go
itsdevbear: agreed.

calbera:
> I think we can move away from KVstore in our own statedb, and use something that propagates the error opposed to panicking, this can simply be a wrapper around kvstore imo. fuck collections tbh.

even if we remove kvstore from statedb, which we should do anyways, the precompiles still use modules that have kvstores

calbera:
so the idea is the _opcodes_ return an error to the _interpreter_ which propogates error up to the EVM for a `VmErr`
similarly, our _kvstores_ should throw errors to the _precompile_interpreter_ (rn its the plugin), to send to EVM

cosmos/lib/evm.go
calbera: ez 
image

itsdevbear: fixed?

itsdevbear:
@calbera is this good or?

cosmos/lib/evm.go
calbera: yes

calbera:
@itsdevbear yeah good to go

calbera:
only the write protection error is caught for panics. all other panics will go up to the currently used handler

calbera:
> @itsdevbear I think the choice is to be clean (finish execution and check for the errors) or consistent with EVM (stop execution immediately and return error).
> 
> * [x]  `ErrExecutionRevert`: we return it as soon as we catch an error so its good
> * [ ]  `ErrOutOfGas`: we wait till the end to check. can use the limited gas meter and panic to halt early
> * [ ]  `ErrWriteProtection`: we can panic on writes/add logs to return immediately or set flags to check at the end. maybe we can use collections in our snap multi store, which returns an error on Set/Delete/etc

in general, i do prefer being consistent with EVM, so I would choose to catch the out of gas error immediately rather than at the end of execution

calbera:
@itsdevbear good to merge imo

itsdevbear:
lgtm

PR #847, opened 2023-06-22
fix: Revert "feat(BlockPlugin): block plugin support of GetHeaderByHash (#…

ocnc:
…814)"

This reverts commit c6e1a8e856c481543cd5705472eeb7f2a4b7ce4f.


itsdevbear:
i think we are okay. 

but definitely do the logger change

PR #874, opened 2023-07-04
fix(docker): Improve docker caching

archbear:
base docker image now copies all go.mod and go.sum files and runs go mod download prior to copying the rest of the go files, caching works as expected

hive mage commands now check for existing base image prior to building, and only builds if necessary + hive now uses the standard base image (no more polaris-hive garbage)


cosmos/docker/base.Dockerfile
```diff
@@ -26,35 +26,10 @@ ARG APP_NAME=polard
 ARG DB_BACKEND=pebbledb
 ARG CMD_PATH=./cosmos/simapp/polard
 ARG FOUNDRY_DIR=contracts
+ARG GO_WORK="contracts eth cosmos magefiles lib e2e/localnet"
```
itsdevbear, line 29: unused?

ocnc (approved):
yeet this into the branch

PR #875, opened 2023-07-05
fix(header): Avoid querying historical versions for storing eth blockhashes

calbera:


cosmos/x/evm/plugins/block/header.go
```diff
@@ -167,3 +171,11 @@ func (p *plugin) writeGenesisHeaderBytes(headerHash common.Hash, headerBz []byte
 func (p *plugin) readGenesisHeaderBytes() []byte {
 	return p.ctx.KVStore(p.storekey).Get([]byte{types.GenesisHeaderKey})
 }
+
+// headerHashKeyForHeight returns the key for the hash of the header at the given height.
+func headerHashKeyForHeight(number int64) []byte {
```
itsdevbear, line 176: put this in a `keys.go` or something

PR #882, opened 2023-07-07
fix(chain): Reduce in-memory cache size

calbera:
the cache size was not actually 64 mb but storing ~67 million items. caches were causing node memory to grow very high. @pentaful1 @BrickBera @transmissions12

### DIFF
--- a/eth/core/chain.go
+++ b/eth/core/chain.go
@@ -35,7 +35,7 @@ import (
 )
 
 // By default we are storing up to 256 items in each cache.
-const defaultCacheSize = 256
+const defaultCacheSize = 1024
 
 // Compile-time check to ensure that `blockchain` implements the `Blockchain` api.
 var _ Blockchain = (*blockchain)(nil)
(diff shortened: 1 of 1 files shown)


eth/core/chain.go
```diff
@@ -34,8 +34,8 @@ import (
 	"pkg.berachain.dev/polaris/eth/params"
 )
 
-// By default we are storing up to 64mb of historical data for each cache.
-const defaultCacheSizeBytes = 1024 * 1024 * 64
+// By default we are storing up to 256 items in each cache.
```
itsdevbear, line 37:
```suggestion
// By default we are storing up to 65536 items in each cache.
```

eth/core/chain.go
```diff
@@ -34,8 +34,8 @@ import (
...
+const defaultCacheSize = 256
```
itsdevbear, line 38:
```suggestion
const defaultCacheSize = 65536
```

itsdevbear (approved):

PR #907, opened 2023-07-17
feat(precompile): Shampoo/plr 104 validate return types

shampoobera:
validate return types for the precompile implementations 

for taking in and returning custom structs, users need to take in and return the geth abigen type for that struct.

TODO: 
- [ ] check input arguments so that we can support overloading
- [ ] add unit tests for wrong function return types/overloading
- [ ] clean up code, it's ugly

lmk wat else.


shampoobera:
nvm im closing this

PR #926, opened 2023-07-28
fix(precompile): fix edge case in precompile matching

shampoobera:


eth/core/precompile/factories.go
```diff
@@ -125,6 +125,9 @@ func buildIdsToMethods(si StatefulImpl, contractImpl reflect.Value) (map[string]
 		if err != nil {
 			return nil, err
 		}
+		if methodName == "" {
```
itsdevbear, line 128: i think this should revert then? 

or do we want to allow it? 

(i guess good for helpers)

calbera: Yea basically it just skips any function that’s not in the ABI, which we should’ve had from the beginning

calbera (approved):
utACK

PR #928, opened 2023-07-30
feat: add genesis builder CLI tool

### LINKED ISSUE #474: refactor(scripts): Refactor the init.sh script into a genesis builder golang cli tool.

anukul:
resolves #474 

added two commands:
- genbuild init moniker: initializes genesis.json, creates 1 account and an associated genesis tx
- genbuild collect: collects genesis txs and updates genesis.json

image


genbuild/cmd/init.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+const (
+	denom          = "abera"
+	keyName        = "dev"
+	initialBalance = 1e26
+	bondAmount     = 1e21
+)
+
+func runInit(cmd *cobra.Command, args []string) {
```
calbera, line 43: this feels like it would be clearer if this can be a init.sh shell script

calbera:
> resolves #474
> 
> added two commands:
> 
> * genbuild init moniker: initializes genesis.json, creates 1 account and an associated genesis tx
> * genbuild collect: collects genesis txs and updates genesis.json
> 
> image

cmd/ should be shell scripts instead of go files to be clearer

anukul:
> cmd/ should be shell scripts instead of go files to be clearer

okay, i probably misinterpreted the issue then.
this is currently in shell [scripts](https://github.com/berachain/polaris/tree/main/e2e/testapp/docker/seed/scripts)

anukul:
felt the need for this while trying to decouple e2e test [fixture](https://github.com/berachain/polaris/blob/64e02d4/cosmos/testing/e2e/precompile/auth/auth_test.go#L47) from containerized node  
for running e2e tests against an in-process node (easier to debug in dev)

this PR was migrating the seed scripts to go as a starting point, but i believe the CLI can be extended to suit more cases (setting up nodes for e2e docker / local, hive etc), which is where shell scripts -> go should make more sense

PR #956, opened 2023-08-09
fix: do something with error returned by stack.Start

technicallyty:
dunno if panic is not preferred here, but i think the error here should be logged one way or another


itsdevbear (approved):
We need a broader fix here, but ty king

PR #990, opened 2023-08-17
fix(mempool): `--slow` fix.

itsdevbear:
The original bug was due to the lock preventing block production and a various weirdness that stemmed from running a validator as an rpc node. Although said use case is discouraged, it causes problems for internal development. This PR adds additional complexity in order to mitigate this issue.


itsdevbear:
fixing tests then g2g

itsdevbear:
@transmissions12 @calbera
https://github.com/berachain/polaris/assets/93690142/f1ef068f-0512-4b3d-abb8-9b54051c0968

itsdevbear:
closing for now

PR #1036, opened 2023-08-28
chore(address): fix conversions 

calbera:


itsdevbear (approved):

cosmos/lib/conversions.go
```diff
@@ -167,18 +168,22 @@ func SdkREToStakingRE(re []stakingtypes.RedelegationEntry) []staking.IStakingMod
...
+		}
 		pubKey, err := val.ConsPubKey()
 		if err != nil {
 			return nil, err
 		}
 		valsOut[i] = staking.IStakingModuleValidator{
-			OperatorAddress: val.OperatorAddress,
-			ConsensusPubkey: pubKey.Bytes(),
+			OperatorAddr:    operEthAddr,
+			ConsAddr:        pubKey.Address(),
```
calbera, line 186: @itsdevbear which of the 2 (`.Address()` or `.Bytes()`) to use here
image

calbera: actually I guess it wouldnt even be using this pubkey but the one in comet

calbera: https://github.com/cometbft/cometbft/blob/main/crypto/sr25519/pubkey.go#L31 Address() is correct, it is already truncated to 20 bytes

calbera: https://github.com/cometbft/cometbft/blob/main/crypto/tmhash/hash.go#L64

itsdevbear:
i dont like but lgtm

PR #1039, opened 2023-08-29
feat(txpool): More optimizations & tests

itsdevbear:


cosmos/x/evm/plugins/txpool/mempool/mempool_reader.go
```diff
@@ -122,6 +123,7 @@ func (etp *EthTxPool) queued() map[common.Address]coretypes.Transactions {
 //
 // NOT THREAD SAFE.
 func (etp *EthTxPool) Nonce(addr common.Address) uint64 {
+	exit := false
 	pendingNonces := make(map[common.Address]uint64)
```
calbera, line 127: there is no need to store a map of addresses actually, we only care about the given `addr`

cosmos/x/evm/plugins/txpool/mempool/mempool_reader.go
```diff
@@ -146,6 +148,10 @@ func (etp *EthTxPool) Nonce(addr common.Address) uint64 {
 			pendingNonces[addr]++
 		case txNonce > pendingNonce+1:
 			// As soon as we see a non contiguous nonce we break.
+			exit = true
+		}
+
+		if exit {
```
calbera, line 154: you can actually use a "label" to do this 
image

cosmos/x/evm/plugins/txpool/mempool/mempool_reader.go
itsdevbear: nah Stats() returns everything

calbera: what? this is just for the nonce function which only cares about the given `addr`

itsdevbear: oh lol ui glitched

itsdevbear:
replaced by #1041

PR #1059, opened 2023-09-04
refactor(pc): string -> [4]bytes

itsdevbear:
- More performant
- Remove `unsafe` and `reflect` usage


itsdevbear:
todo: write a quick test to make sure it's not using the pointer memory address as the key by accident

eth/core/precompile/stateful_container.go
```diff
@@ -42,7 +41,7 @@ type statefulContainer struct {
 	// method signatures) to native precompile functions. The signature key is provided by the
 	// precompile creator and must exactly match the signature in the geth abi.Method.Sig field
 	// (geth abi format). Please check core/precompile/container/method.go for more information.
-	idsToMethods map[string]*method
+	idsToMethods map[[4]byte]*method
```
calbera, line 44: i would make a `type methodSelector [4]byte` or smth

calbera: `type methodSelector [NumBytesMethodID]byte`

eth/core/precompile/method.go
```diff
@@ -31,6 +31,9 @@ import (
 	"pkg.berachain.dev/polaris/lib/utils"
 )
 
+// methodID is a fixed length byte array that represents the method ID of a precompile method.
+type methodID = [NumBytesMethodID]byte
```
calbera, line 35: nit but I think this should be a type alias rather than a type declaration

itsdevbear: sure

calbera (approved):

PR #1069, opened 2023-09-05
feat(distr): add getWithdrawAddress

calbera:


contracts/src/cosmos/precompile/Distribution.sol
```diff
@@ -34,6 +34,8 @@ interface IDistributionModule {
 
     function getWithdrawEnabled() external view returns (bool);
 
+    function getWithdrawAddress(address delegtor) external view returns (address);
```
itsdevbear, line 37: spelling also add natspec

ocnc (approved):

transmissions12 (approved):
utAck

PR #1076, opened 2023-09-07
fix(pc): Enforce read-only at the container level

calbera:
Snapshot before and revertToSnapshot after execution of view/pure method; can remove cacheContexts

### DIFF
--- a/eth/core/precompile/stateful_container.go
+++ b/eth/core/precompile/stateful_container.go
@@ -83,7 +83,8 @@ func (sc *statefulContainer) Run(
 	}
 
 	// If the method is read-only (view/pure), snapshot so any state changes made during the call
-	// are not persisted.
+	// are not persisted. NOTE: this may not be the best place to handle state mutability, but it
+	// gets the job done.
 	if method.abiMethod.IsConstant() {
 		sdb := evm.GetStateDB()
 		snapshot := sdb.Snapshot()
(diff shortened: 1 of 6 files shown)


itsdevbear (approved):
lgtm, I'd add a test tho

ocnc (approved):
lgtm

itsdevbear:
fix lint then g2g

eth/core/precompile/stateful_container.go
```diff
@@ -82,6 +82,14 @@ func (sc *statefulContainer) Run(
 		return nil, ErrMethodNotFound
 	}
 
+	// If the method is read-only (view/pure), snapshot so any state changes made during the call
+	// are not persisted.
+	if method.abiMethod.IsConstant() {
```
itsdevbear, line 87: feels dirty to use the snapshot feature at the container level ngl. 

the container should just run, it doesn't have any knowledge of snapshots. 

Can we put it higher up? 

if not just put a note saying this is not good separation of concerns but it is what it is for now

calbera: the way this is designed (and maybe needs to be restructured) is 1) EVM --> 2) PrecompileMgr (cosmos pc plugin) -->  3) StatefulContainer (polaris eth). 1) sets EVM requirements. 2) sets cosmos specific rules but has no idea of what the precompile will execute. 3) only has access to the abi method, which can be checked for state-mutability, payability, etc.

calbera: in theory we could just expose the methods on the container so that the cosmos layer can handle all control of state modifications.

ocnc (approved):

PR #1080, opened 2023-09-07
fix(txpool): get state from disk from nonce

calbera:


itsdevbear (dismissed):
add comment with todo other than that looks good to me

ocnc (approved):
anotha LGTM

PR #1081, opened 2023-09-07
feat: upgrade Rollkit x Polaris "rollkit" branch

jcstein:
this is to update the "rollkit" branch of berachain/polaris, ahead of adding this to the Rollkit documentation.


jcstein:
hrm, closing this as there are some old op scripts in here that i don't think are necessary for this PR

PR #1082, opened 2023-09-07
fix(mempool): Re-add locks

itsdevbear:


transmissions12 (approved):
Good w this interim fix. Need to improve mempool design

PR #1096, opened 2023-09-13
feat(infra): Support `sync_progress` and reduce spammy tx logs.

itsdevbear:
- Help brick manage node status better
- Reduce spammy logs which I think may be a non trivial I/O bottleneck


itsdevbear:
should be drop in to `berachain`

calbera (approved):
utACK

transmissions12 (approved):
Lgtm

PR #1101, opened 2023-09-18
refactor(node): Deprecate `graphql`

itsdevbear:
No-one uses this fr and its utilizing resources and causing flaky af tests.


ocnc (approved):
nuke it

itsdevbear:
i'm in simplify as much as possible mode rn

PR #1105, opened 2023-09-18
refactor(eth): `txpool` breakout

itsdevbear:


eth/polar/mining.go
```diff
@@ -30,7 +30,13 @@ import (
...
 // Prepare prepares the Polaris chain for processing a new block at the given height.
 func (pl *Polaris) Prepare(ctx context.Context, number uint64) {
-	pl.blockchain.Prepare(ctx, number)
+	header := pl.blockchain.Prepare(ctx, number)
+	if header == nil {
+		panic("blockchain produced nil header")
+	}
+	// We update the base fee in the txpool to the next base fee.
+	// TODO: Move to prepare proposal
+	pl.txPool.SetBaseFee(header.BaseFee)
```
calbera, line 39: this should happen in the previous block's endblock/precommit

itsdevbear: that will break consensus :(

itsdevbear: needs to be done in process / prepare proposal to be 100% correct. I think lets push off changing anything until I get there.

itsdevbear: also probs gonna be addressed in #1107

ocnc (approved):

PR #1113, opened 2023-09-19
fix(stack): decrease stack allocs

calbera:


itsdevbear:
@calbera tbh we should use a `sync.Pool`

https://github.com/ethereum/go-ethereum/blob/master/core/vm/stack.go

itsdevbear:
otherwise lgtm

itsdevbear (approved):
add sync pool

itsdevbear:
later ting

PR #1129, opened 2023-09-20
fix(evm): missing things when init or export genesis

taeguk:
There are some missing handlings for nonce and balance.


itsdevbear (approved):

itsdevbear:
can you add export n import tests?

itsdevbear:
ended up actually reverting this, looks like its causing HIVE test to fail,

itsdevbear:
@calbera can you take a look today?

PR #1131, opened 2023-09-20
refactor(abci): `miner` phase 2 - custom prepare/process proposal

itsdevbear:


cosmos/abci/prepare/default.go
```diff
@@ -0,0 +1,126 @@
...
+		maxBlockGas = b.MaxGas
+	}
+
+	var (
+		selectedTxs  [][]byte
+		totalTxBytes int64
+		totalTxGas   uint64
+	)
+
+	iterator := h.mempool.Select(ctx, req.Txs)
```
itsdevbear, line 74: instead of calling `mempool.Select()` here we should call some sort of variation on `miner.generateWork` 
https://github.com/ethereum/go-ethereum/blob/5b9cbe30f8ca2487c8991e50e9c939d5e6ec3cc2/miner/worker.go#L1005

PR #1136, opened 2023-09-24
fix(precompile): Fix regression where `eth_call` can cause non-deterministic gas consumption

itsdevbear:
You cannot manipulate the canonical state plugin in another plugin since there is no guarantee made about whether it is in a query or a block. Also setting gas consumption in the state plugin is not required, since the context with the updated gas only occurs in the precompile itself.

This can cause non-determinism in gas consumption between nodes.


ocnc (approved):

cosmos/x/evm/plugins/precompile/plugin.go
```diff
@@ -185,9 +182,6 @@ func (p *plugin) enableReentrancy(sdb vm.PolarisStateDB) {
 	// end precompile execution => stop emitting Cosmos event as Eth logs for now
 	cem := utils.MustGetAs[state.ControllableEventManager](sdkCtx.EventManager())
 	cem.EndPrecompileExecution()
-
-	// remove Cosmos gas consumption so gas is consumed only per OPCODE
-	p.sp.SetGasConfig(storetypes.GasConfig{}, storetypes.GasConfig{})
```
calbera, line 190: Can we not just 0 the gas config on the given statedb here, would make much more sense

cosmos/x/evm/plugins/precompile/plugin.go
```diff
@@ -203,7 +197,4 @@ func (p *plugin) disableReentrancy(sdb vm.PolarisStateDB) {
 	// resume precompile execution => begin emitting Cosmos event as Eth logs again
 	cem := utils.MustGetAs[state.ControllableEventManager](sdkCtx.EventManager())
 	cem.BeginPrecompileExecution(sdb)
-
-	// restore ctx gas configs for continuing precompile execution
-	p.sp.SetGasConfig(p.kvGasConfig, p.transientKVGasConfig)
```
calbera, line 208: Here as well

itsdevbear:
fwiw, I really hate this design, but it will get us unblocked for testnet

itsdevbear:
We can chat monday, but I really think we should just simplify and prevent precompiles from calling back into the EVM. This would reduce so much complexity fromt the precompile plugin and stop just gross interface contagion. I think the fact that the precompile plugin is manually calling the state plugin is a huge antipattern and generally bad practice. Plugins in general, shouldnt need to know about the existance of other plugins.

itsdevbear:
@calbera @ocnc

transmissions12 (approved):
Lgtm thanks for addressing it @itsdevbear

PR #1156, opened 2023-09-27
fix(cosmos): Deterministically load genesis accounts into StateDB

calbera:
### Summary by CodeRabbit

- Refactor: Improved consistency in the `InitGenesis` function by sorting addresses in ascending order.
- New Feature: Added a new type `SortableAddresses` for sorting Ethereum addresses based on their hexadecimal representation.
- Bug Fix: Enhanced error handling in various functions including `InitGenesis`, `SetChainConfig`, and others, preventing potential panics and allowing errors to be handled more gracefully.
- Test: Introduced new test cases to verify the correct sorting of addresses and successful initialization of genesis state.
- Refactor: Updated `InitGenesis` function to check account nonce consistency with the auth keeper, improving security.
- Bug Fix: Fixed an issue where accounts with code did not have a minimum nonce of 1 as per EIP-161 requirements.


eth/core/genesis.go
```diff
@@ -76,3 +76,10 @@ func UnmarshalGenesisHeader(header *types.Header, gen *Genesis) {
 	gen.Mixhash = header.MixDigest
 	gen.Coinbase = header.Coinbase
 }
+
+// SortableAddresses is a sortable slice of common.Addresses.
+type SortableAddresses []common.Address
+
+func (a SortableAddresses) Len() int           { return len(a) }
```
itsdevbear, line 83: maybe add something that prevents duplicates? so that that doesn't introduce non-determinism / a bad genesis

cosmos/x/evm/plugins/state/genesis.go
```diff
@@ -22,24 +22,34 @@ package state
...
 		// TODO: technically wrong since its overriding / hacking the auth keeper and
 		// we are using the nonce from the account keeper as well.
+		account := ethGen.Alloc[address]
 		p.CreateAccount(address)
-		p.SetBalance(address, account.Balance)
+		if account.Balance != nil {
+			p.SetBalance(address, account.Balance)
+		}
 		if account.Code != nil {
 			p.SetCode(address, account.Code)
```
itsdevbear, line 54: add check to make sure if code isn't nil nonce >= 1 if the proper hard fork is enable (i forget the exact one)

itsdevbear: also add a check that nonce == Statedb.GetNonce() and then add note that shows that because account keeper is earlier in InitGenesisOrder, the noncing here is redundant

eth/core/genesis.go
calbera: actually will panic earlier as the json struct is being unmarshalled into the golang genesisAccounts map (keyed on addresses)

itsdevbear: does it have a check for it?

cosmos/x/evm/plugins/state/genesis.go
calbera: its eip 161 which was released with eip155 (theres no check for 161 specifically)

itsdevbear: check `EIP155Block`

itsdevbear: i think geth has a VerifyGenesis function

calbera: well by default, its 2million smth but it can be modified in the given genesis to be 0, in which case the nonce must be >= 1 for an account with code

eth/core/genesis.go
calbera: geth has no `VerifyGenesis`. It just overwrites it if given multiple accounts with the same address in the map.That behavior makes sense imo

cosmos/x/evm/plugins/state/genesis.go
calbera: actually the geth default is 2mil smth and polaris default is 0

eth/core/genesis.go
calbera: just checked "When unmarshaling a JSON object with duplicate keys into a Go map using json.Unmarshal(), the last value for the duplicate key will be used"

itsdevbear: this could cause non determinism if the order for which sort sorts the two duplicate keys is non deterministic, we should explicitly check for this case and panic with a duplicate key error message

itsdevbear:
@calbera or write a validate basic that checks for duplicate keys

calbera:
@itsdevbear 
To detect duplicate keys in raw JSON data before unmarshaling it into a Go map, you can use the `json.Decoder` to walk through the JSON and check for duplicate keys[1]. Here's an example of how to do this:

```go
package main

import (
	"bytes"
	"encoding/json"
	"fmt"
	"io"
)

func main() {
	jsonData := []byte(`{"key": "value1", "key": "value2"}`)

	duplicates, err := findDuplicateKeys(jsonData)
	if err != nil {
		fmt.Println("Error:", err)
		return
	}

	if len(duplicates) > 0 {
		fmt.Println("Duplicate keys found:", duplicates)
	} else {
		fmt.Println("No duplicate keys found")
	}
}

func findDuplicateKeys(jsonData []byte) ([]string, error) {
	dec := json.NewDecoder(bytes.NewReader(jsonData))
	var duplicates []string

	for {
		t, err := dec.Token()
		if err == io.EOF {
			break
		}
		if err != nil {
			return nil, err
		}

		if dec.More() {
			if t, ok := t.(string); ok {
				for _, d := range duplicates {
					if d == t {
						return duplicates, fmt.Errorf("duplicate key found: %s", t)
					}
				}
				duplicates = append(duplicates, t)
			}
		}
	}

	return duplicates, nil
}
```

In this example, the `findDuplicateKeys` function takes raw JSON data as input and uses a `json.Decoder` to iterate through the JSON tokens. When it encounters a string token, it checks if the token is already in the `duplicates` slice. If it is, an error is returned with the duplicate key. If not, the token is added to the `duplicates` slice.

When running this code with the given JSON data containing duplicate keys, the output will be:

```
Duplicate keys found: [key]
```

This approach allows you to detect duplicate keys in raw JSON data before unmarshaling it into a Go map. Note that this example assumes a flat JSON object structure. If you have nested objects, you'll need to modify the function to handle them accordingly.

Citations:
[1] https://stackoverflow.com/questions/50107569/detect-duplicate-in-json-string-golang
[2] https://github.com/golang/go/issues/24415
[3] https://pkg.go.dev/gopkg.in/square/go-jose.v2/json
[4] https://groups.google.com/g/golang-dev/c/zBQwhm3VfvU
[5] https://golang-nuts.narkive.com/wzsECw7c/go-nuts-encoding-json-surprising-behaviour-when-marshalling-duplicate-keys
[6] https://www.digitalocean.com/community/tutorials/how-to-use-json-in-go

eth/core/genesis.go
calbera: there wouldn't be any non-determinism as the json file is read in order, so always the last of the duplicates will be included in the map

cosmos/x/evm/plugins/state/genesis.go
```diff
@@ -21,38 +21,69 @@
...
+		} else if account.Nonce != 0 {
+			p.SetNonce(address, account.Nonce)
+		}
+
+		// initialize the account data on the state plugin
+		if account.Balance != nil {
+			p.SetBalance(address, account.Balance)
+		}
 		if account.Code != nil {
+			if ethGen.Config.IsEIP155(big.NewInt(0)) && account.Nonce == 0 {
```
itsdevbear, line 68: in my pr this will be killed so maybe handle it another way pre-emptively

itsdevbear:
Is your sort function stable tho? The sort could change the order is what
i’m saying

itsdevbear:
https://www.geeksforgeeks.org/stable-and-unstable-sorting-algorithms/amp/

calbera:
Sure, can use a stable sort algorithm, but it shouldnt matter since the slice is guaranteed to have no duplicate addresses

cosmos/x/evm/plugins/state/genesis.go
itsdevbear: yeah cause its using the one from actual live mainnet chainid 1

cosmos/x/evm/plugins/state/genesis.go
calbera: sure moving to validate genesis

itsdevbear (approved):

PR #1161, opened 2023-09-29
feat(sdb): Return  `[]byte{0x01}` for `GetCode(..)` of precompile contracts

itsdevbear:
### Summary by CodeRabbit

- New Feature: Introduced a `PrecompilePlugin` interface and integrated it into various components such as the state database, miner, and host setup. This allows for more flexible handling of precompiled contracts.
- Refactor: Updated the initialization of `precompile.NewPlugin` and `state.NewPlugin` instances in the `host.go` file to ensure correct setup based on the current block's rules.
- Test: Added generated mock packages for testing purposes, including mocks for `Accesslist`, `Log`, `Refund`, `SelfDestructStatePlugin`, `SelfDestructs`, `TransientStorage`, `Plugin`, and `PrecompilePlugin`.
- Chore: Updated `.mockery.yaml` configuration settings for two packages to include all files recursively and set the `with-expecter` flag to true.


eth/core/processor.go
```diff
@@ -87,12 +87,6 @@ func NewStateProcessor(
 		statedb:  statedb,
 	}
 
-	if sp.pp == nil {
-		sp.pp = precompile.NewDefaultPlugin()
-	} else {
-		sp.BuildAndRegisterPrecompiles(sp.pp.GetPrecompiles(nil))
-	}
```
calbera, line 94: still need this

eth/core/processor.go
```diff
@@ -124,6 +118,7 @@ func (sp *StateProcessor) Prepare(evm *vm.GethEVM, header *types.Header) {
 	// *technically* the precompiles change based on the chain config rules, to be fully correct,
 	// we should check every block.
 	sp.BuildAndRegisterPrecompiles(precompile.GetDefaultPrecompiles(&rules))
+	sp.BuildAndRegisterPrecompiles(sp.pp.GetPrecompiles(&rules))
```
calbera, line 121: only need to re-register the geth stateless precompiles every block

eth/core/processor.go
itsdevbear: moved it to the miner.

eth/core/processor.go
itsdevbear: will be needed (at least for now), in my geth tx pool branch

calbera (approved):

PR #1214, opened 2023-10-10
refactor(ante): Remove ante handler as it is not needed

itsdevbear:
### Summary by CodeRabbit

"Refactor: Removed Custom Ante Handler

- The custom Ante Handler, previously imported from the `antelib` package, has been removed from the application. This change simplifies the codebase and reduces potential points of failure. Note that this may affect how transactions are processed if you were relying on the custom Ante Handler's functionality."


e2e/testapp/app.go
```diff
@@ -213,8 +212,8 @@ func NewPolarisApp(
 	app.mm = miner.New(app.EVMKeeper.Polaris().Miner())
 	app.SetPrepareProposal(app.mm.PrepareProposal)
 
-	// Setup Custom Ante Handler
-	app.SetAnteHandler(antelib.NewMinimalHandler())
+	// Set the ante handler to nil, since it is not needed.
```
calbera, line 215: well said

calbera (approved):

PR #1240, opened 2023-10-21
refactor(statedb): Remove account keeper from state plugin

itsdevbear:
### Summary by CodeRabbit

- Refactor: Simplified the codebase by removing the dependency on `AccountKeeper` across multiple files. This change streamlines the code and reduces complexity.
- New Feature: Added a function `NonceKeyFor` to generate a unique key for each address, enhancing data organization and retrieval.
- Refactor: Updated the `InitGenesis` function to unconditionally set the nonce value for each account, simplifying the logic.
- Test: Enhanced JSON RPC functionality tests with additional checks for nonce values and a new helper function `ExpectSuccessReceipt`, improving the robustness of our testing suite.
- Refactor: Updated the `plugin` struct to directly access the key-value store, removing the need for `AccountKeeper`. This change improves the efficiency of account-related operations.
  
These changes primarily focus on codebase simplification and testing enhancements, improving maintainability and reliability of the software.


itsdevbear:
looks like there do be a gremlin in here somewhere

itsdevbear:
surrounding pending nonce

cosmos/x/evm/plugins/state/plugin.go
```diff
@@ -331,11 +302,6 @@ func (p *plugin) SetNonce(addr common.Address, nonce uint64) {
 // GetCodeHash implements the `StatePlugin` interface by returning
 // the code hash of account.
 func (p *plugin) GetCodeHash(addr common.Address) common.Hash {
-	if !p.ak.HasAccount(p.ctx, addr[:]) {
-		// if account at addr does not exist, return zeros
-		return common.Hash{}
-	}
```
calbera, line 337: can still keep this check with the NonceKey no?

itsdevbear: It's actually incorrect, i also realized that 32 empty bytes and 32 filled bytes allocates 32 bytes regardless so might as well return empty code hash (0x5c...)

calbera: even though we're allocating memory, if the EVM differentiates between {0000...0} and (0x5c...) code hashes, then should be careful about it

itsdevbear: Hmmm.....

```
// from geth https://github.com/ethereum/go-ethereum/blob/master/core/vm/instructions.go#L418
func opExtCodeHash(pc *uint64, interpreter *EVMInterpreter, scope *ScopeContext) ([]byte, error) {
	slot := scope.Stack.peek()
	address := common.Address(slot.Bytes20())
	if interpreter.evm.StateDB.Empty(address) {
		slot.Clear()
	} else {
		slot.SetBytes(interpreter.evm.StateDB.GetCodeHash(address).Bytes())
	}
	return nil, nil
}
```

https://eips.ethereum.org/EIPS/eip-1052#specification

image

itsdevbear:
```
// CreateAccount implements the `StatePlugin` interface by creating a new account
// in the account keeper. It will allow accounts to be overridden.
func (p *plugin) CreateAccount(addr common.Address) {
	store := p.cms.GetKVStore(p.storeKey)
	store.Set(NonceKeyFor(addr), sdk.Uint64ToBigEndian(0))
	// initialize the code hash to empty
	store.Set(CodeHashKeyFor(addr), emptyCodeHashBytes)
}
```

accounts that don't exist will have 0 codehash, accounts that exist with get set to 0xc5d24....

itsdevbear: then our heuristic for if a an account exists is whether our nonce is 0 or nil.

PR #1266, opened 2023-11-01
refactor(blockchain): Minor cleanup and add safety checks.

itsdevbear:
## Summary by CodeRabbit

Refactor:
- Enhanced error messaging in the `Precommit` function for better debugging.
- Updated the `InitGenesis` function for improved block processing.
- Modified the `Blockchain` interface and `blockchain` struct for better block handling and validation.
- Improved block insertion and processing logic in the `ChainWriter` interface.

New Features:
- Added `HasBlockAndState` function to check the existence of a block and its state.
- Introduced `validator` field in the `blockchain` struct for block validation.

Bug Fixes:
- Removed a debug log statement from the `SubscribeLogsEvent` function for cleaner logs.

Tests:
- Increased the number of blocks to wait for in the "Distribution Precompile" test for more reliable results.

Chores:
- Updated import statements and added new methods to interfaces for better code organization and functionality.

### DIFF
--- a/cosmos/miner/miner.go
+++ b/cosmos/miner/miner.go
@@ -47,7 +47,8 @@ type EnvelopeSerializer interface {
 }
 
 type App interface {
-	BeginBlocker(ctx sdk.Context) (sdk.BeginBlock, error)
+	BeginBlocker(sdk.Context) (sdk.BeginBlock, error)
+	PreBlocker(sdk.Context, *abci.RequestFinalizeBlock) (*sdk.ResponsePreBlock, error)
 }
 
 // EVMKeeper is an interface that defines the methods needed for the EVM setup.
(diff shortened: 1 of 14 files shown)


eth/core/chain_writer.go
```diff
@@ -22,83 +22,160 @@ package core
...
+	return err
+}
 
-	// *********************************************** //
-	// TODO: restructure this function / flow it sucks //
-	// *********************************************** //
-	blockHash, blockNum := block.Hash(), block.Number().Uint64()
-	bc.logger.Info(
-		"finalizing evm block", "hash", blockHash.Hex(), "num_txs", len(receipts))
+// SetHeadAndFinalize sets the head of the blockchain to the given block and finalizes the block.
```
calbera, line 85:
```suggestion
// WriteBlockAndSetHead sets the head of the blockchain to the given block and finalizes the block.
```

ocnc (approved):

calbera (approved):

cosmos/miner/miner.go
```diff
@@ -68,11 +85,34 @@ func (m *Miner) Init(serializer EnvelopeSerializer) {
 func (m *Miner) PrepareProposal(
```
itsdevbear, line 85: @alexanderbez is this safe?

alexanderbez: I haven't dug into the specifics of each individual method call, but since `PrepareProposal` state is discarded upon round timeout, any state modifications should be safe assuming you use the provided `ctx` (which it looks like you are!).

alexanderbez: If you'd like, I can dive into the individual method calls for further analysis.

PR #1302, opened 2023-11-08
feat(oe): Enable Optimistic Execution

itsdevbear:
IMO from reading the baseapp code I see no reason why not to enable this.


tac0turtle:
> @tac0turtle is this is the roadmap?

yes it will be easy to do this with server/v2. We are trying to get it working this week

itsdevbear:
@tac0turtle sweet lmk and I can test it out, it should help imporve our performance a ton.

PR #1315, opened 2023-11-16
fix(config): update genesis state

BrickBera:
## Summary by CodeRabbit

- **New Features**
  - Integrated a new default chain configuration for the genesis block.
- **Refactor**
  - Updated the allocation of balances within the genesis block to reflect new parameters.

### DIFF
--- a/eth/core/genesis.go
+++ b/eth/core/genesis.go
@@ -26,6 +26,7 @@ import (
 	"github.com/berachain/polaris/eth/common"
 	"github.com/berachain/polaris/eth/common/hexutil"
 	"github.com/berachain/polaris/eth/core/types"
+	"github.com/berachain/polaris/eth/params"
 
 	"github.com/ethereum/go-ethereum/core"
 )
@@ -38,6 +39,9 @@ type (
 
 // DefaultGenesis is the default genesis block used by Polaris.
 var DefaultGenesis = &core.Genesis{
+	// Chain Config
+	Config: params.DefaultChainConfig,
+
 	// Genesis Block
 	Nonce:     0,
 	Timestamp: 0,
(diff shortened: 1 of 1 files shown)


eth/core/genesis.go
```diff
@@ -50,9 +54,30 @@ var DefaultGenesis = &core.Genesis{
 
 	// Genesis Accounts
 	Alloc: core.GenesisAlloc{
-		// 0xfffdbb37105441e14b0ee6330d855d8504ff39e705c3afa8f859ac9865f99306
-		common.HexToAddress("0x20f33CE90A13a4b5E7697E3544c3083B8F8A51D4"): {
-			Balance: big.NewInt(0).Mul(big.NewInt(5e18), big.NewInt(100)), //nolint:gomnd // its okay.
+		// 0xbac
+		common.HexToAddress("0xFE94Cc9f0dfbb657a6C6850701aBF6356227F8c3"): {
+			Balance: big.NewInt(0).Mul(big.NewInt(1e18), big.NewInt(1e7)), //nolint:gomnd // its okay.
```
technicallyty, line 59: would suggest doing:

```go
var (
    coins = big.NewInt(0).Mul(big.NewInt(1e18), big.NewInt(1e7))
)
...
		common.HexToAddress("0xFE94Cc9f0dfbb657a6C6850701aBF6356227F8c3"): {
			Balance: coins,
```
and pass it into each balance so:

1. you don't have to do the nolint line so many times
2. easily configurable if you ever need to change all these balances in one swoop

itsdevbear: @BrickBera the Config: isn't used in genesis atm, also is this change for devnet?

BrickBera: > @BrickBera the Config: isn't used in genesis atm, also is this change for devnet?

Need the config to change evm chain-id. The balances change is for future devnet redeploys

itsdevbear:
should be done via configuration of the node, not via changing the default config

PR #1320, opened 2023-11-20
fix(e2e): pin polaris & rollkit versions to testapp

technicallyty:
pins polaris and rollkit versions to go.mod to ensure the tesapp can run in isolation outside of go.work


jcstein (approved):
:shipit:

PR #1341, opened 2023-12-06
chore(distribution): Test calling query view methods in distribution precompile

calbera:
- Related to issue of calling view functions which have underlying cosmos behavior that makes state changes


itsdevbear:
Is this just adding a test?

ocnc (approved):

Issue #547, opened 2023-04-13
feat(cosmos): add pagination to Cosmos requests in precompiles 


BrickBera:
in precompiles, e.g. bank precompile, there are Cosmos requests that can take pagination as a request param:
e.g.
https://github.com/berachain/polaris/blob/26526ee6c28d5c0fdf00a26bf4b5acb8507c1de9/cosmos/precompile/bank/bank.go#L245

PR #1344, opened 2023-12-08
feat(cosmos): Issue#547 adding pagination to cosmos precompile

j-ard:
I'm seeking to add pagination to any of the cosmos precompile modules that support it, starting with the one explicitly called out in [Issue#547](https://github.com/berachain/polaris/issues/547). 

Currently the project builds and passes unit tests. 

I'd be happy for the opportunity to write some test improvements as well, as they are a bit lacking in the DRY department.

Thanks in advance for your review!


calbera:
Thanks for the PR!

PR #1392, opened 2024-01-05
Expose some GetStateDB in PolarContext

mycodecrafting:
- Adds precompile address to PolarContext
- Adds read / write functionality onto the PolarContext:
  - AddBalance
  - SubBalance 
  - SetState
  - GetState

### DIFF
--- a/cosmos/precompile/bank/bank_test.go
+++ b/cosmos/precompile/bank/bank_test.go
@@ -140,7 +140,7 @@ var _ = Describe("Bank Precompile Test", func() {
 		)
 
 		BeforeEach(func() {
-			ctx = vm.NewPolarContext(ctx, nil, common.BytesToAddress(acc), big.NewInt(0))
+			ctx = vm.NewPolarContext(ctx, nil, common.Address{}, common.BytesToAddress(acc), big.NewInt(0))
 		})
 
 		When("GetBalance", func() {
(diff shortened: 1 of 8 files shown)


calbera:
Hey @mycodecrafting thanks for the PR! You can currently get the `StateDB` off of the current `PolarContext`. In a precompile, it could look something like (for a given `ctx context.Context`):

```
import (
  pvm "github.com/berachain/polaris/eth/core/vm"
)

polarCtx := vm.UnrwapPolarContext(ctx)
stateDB := polarCtx.Evm().GetStateDB()
```

mycodecrafting:
> Hey @mycodecrafting thanks for the PR! You can currently get the `StateDB` off of the current `PolarContext`. In a precompile, it could look something like (for a given `ctx context.Context`):
> 
> ```
> import (
>     "github.com/berachain/polaris/eth/core/vm"
> )
> 
> polarCtx := vm.UnwrapPolarContext(ctx)
> stateDB := polarCtx.Evm().GetStateDB()
> ```

Hey, yes I am aware of this. The PR is a result of a convo with @itsdevbear a bit back around work that we're doing on our fork of Polaris. The reason for the change is to make precompile state management easier, and a bit safer. E.g, when  using the proposed SetState GetState on the Context, the precompile dev cannot accidentally write to the storage for other addresses.

Apologies if I misunderstood and this is not an appropriate upstream change.

PR #1418, opened 2024-01-14
fix(config): fix validator config params

itsdevbear:


cosmos/runtime/txpool/mempool.go
```diff
@@ -125,28 +144,55 @@ func (m *Mempool) Insert(ctx context.Context, sdkTx sdk.Tx) error {
...
-		[]*ethtypes.Transaction{wet.Unwrap()}, false, false,
-	); len(errs) != 0 {
-		// Handle case where a node broadcasts to itself, we don't want it to fail CheckTx.
-		if errors.Is(errs[0], ethtxpool.ErrAlreadyKnown) && sCtx.ExecMode() == sdk.ExecModeCheck {
-			return nil
+// processInserts processes inserts into the txpool.
+// this is basically used to process remote transactions
+// incoming from comet.
+func (m *Mempool) processInserts() {
+	m.wg.Add(1)
```
nivasan1, line 161: nit: generally you call wg.Add from the main GR (the GR that blocks on wait), this is to avoid potential race-conditions between concurrent Add / Wait calls

cosmos/runtime/txpool/mempool.go
```diff
@@ -125,28 +144,55 @@ func (m *Mempool) Insert(ctx context.Context, sdkTx sdk.Tx) error {
...
+	ticker := time.NewTicker(insertTicker)
+	defer ticker.Stop() // stop ticker
+	for {
+		select {
+		case <-m.stopInsertCh:
+			return
+		case <-ticker.C:
+			// We treat this as a "read", because we want
+			// prepare proposal to always win thie lock contention here.
+			m.safetyLock.RLock()
```
nivasan1, line 174: nit: why only take a read-lock here? Is txpool.Add not mutating the txpool?

cosmos/runtime/miner/miner.go
```diff
@@ -73,6 +80,9 @@ func (m *Miner) Init(serializer EnvelopeSerializer) {
 func (m *Miner) buildBlock(ctx sdk.Context) ([]byte, uint64, error) {
 	defer m.clearPayload()
 
+	m.txPoolSafety.Lock()
```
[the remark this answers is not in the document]
nivasan1: edit: looks like this is to avoid concurrent aditions to the tx-pool while `buildBlock` is being called?

PR #1419, opened 2024-01-14
fix(val): val config1

itsdevbear:
## Summary by CodeRabbit

- **New Features**
  - Introduced recommended configuration functions for various system components.
  - Enhanced miner data initialization with commit information.
  - Updated transaction status logic for improved clarity and performance.
  - Added new transaction queuing system to handle transactions post-block commitment.

- **Bug Fixes**
  - Fixed an issue where transactions might not be correctly ejected from the mempool under certain conditions.
  - Corrected the return type of the `EthTransaction` function to align with updated protocol specifications.

- **Refactor**
  - Simplified configuration settings for testing environments.
  - Updated the handling of transaction broadcasting in the transaction pool.

- **Documentation**
  - Updated code generation tool version comments across multiple mock files.

- **Style**
  - Standardized chain ID naming conventions and initialization blocks for better readability.

- **Chores**
  - Removed outdated code and unused functions from various test and state management files.

- **Tests**
  - Adjusted EVM genesis block tests to use the new default configuration settings.

- **Revert**
  - No revert actions in this release.


cosmos/runtime/txpool/ante.go
```diff
@@ -46,6 +45,7 @@ func (m *Mempool) AnteHandle(
 	if ctx.ExecMode() == sdk.ExecModeReCheck {
 		if wet, ok := utils.GetAs[*types.WrappedEthereumTransaction](msgs[0]); ok {
 			if m.shouldEjectFromCometMempool(ctx.BlockTime(), wet.Unwrap()) {
+				delete(m.receivedFromCometAt, wet.Unwrap().Hash())
```
calbera, line 48: lock on deletion

cosmos/runtime/txpool/mempool.go
```diff
@@ -104,10 +111,30 @@ func (m *Mempool) Insert(ctx context.Context, sdkTx sdk.Tx) error {
...
+	if !ok {
 		// We have to return nil for non-ethereum transactions as to not fail check-tx.
 		return nil
-	} else if errs := m.txpool.Add(
+	}
+
+	// If we already have the transaction in the txpool, we can return early.
+	ethTx := wet.Unwrap()
+	if ethTx == nil {
+		return errors.New("wraped tx is nil")
```
cyka-bera, line 123: wrapped

cosmos/runtime/txpool/ante.go
```diff
@@ -60,27 +62,24 @@ func (m *Mempool) shouldEjectFromCometMempool(
...
-		currentTime.Sub(tx.Time()) > m.lifetime
+	// 2. If the transaction has been in the mempool for longer
+	// than the configured timeout. (txs view).
+	// 3, If the transaction has been in the mempool for longer
+	// than the configured timeout. (our view).
+	m.receivedFromCometAtMu.RLock()
+	cometTime := m.receivedFromCometAt[txHash]
+	m.receivedFromCometAtMu.RUnlock()
+	return m.inCanonicalChain(txHash) ||
+		currentTime.Sub(tx.Time()) > m.lifetime ||
```
calbera, line 77: just double checked `tx.Time()` is set on line 47 in `wet.Unwrap()` (because a new Transaction object gets created and `time` field is set in `UnmarshalBinary`). `currentTime` is `ctx.BlockTime()`, which will actually be earlier so this case will always be false

cosmos/runtime/txpool/mempool.go
```diff
@@ -104,10 +113,30 @@ func (m *Mempool) Insert(ctx context.Context, sdkTx sdk.Tx) error {
...
+	}
+
+	// If the tx is a local, or has been gossiped again for some reason. We ignore it.
+	ethTxHash := ethTx.Hash()
+	if m.txpool.Has(ethTxHash) {
+		return nil
+	}
```
calbera, line 132: Has takes the read lock on txpool, so will block miner from building block (both `Pending` and `Locals` do a full RW lock in worker building block). In stage 1 its fine bc both this call and `txpool.Add` lock on mempool. but in stage2 we fully optimize by doing the `pauseInserts`, which means removing the `.Has` check

cosmos/config/default.go
```diff
@@ -0,0 +1,72 @@
...
+	cfg.TxIndex.Indexer = "null"
+
+	cfg.Instrumentation.Prometheus = true
+	return cfg
+}
+
+// RecommendedServerConfig returns the recommended server config.
+func RecommendedServerConfig() *serverconfig.Config {
+	cfg := serverconfig.DefaultConfig()
+	cfg.MinGasPrices = "0abera"
```
nivasan1, line 53: It'd probs make sense to increment `MinGasPrices` to avoid potential spam txs from entering the mempl?

cosmos/runtime/txpool/ante.go
```diff
@@ -43,9 +42,12 @@ func (m *Mempool) AnteHandle(
 	msgs := tx.GetMsgs()
 
 	// We only want to eject transactions from comet on recheck.
-	if ctx.ExecMode() == sdk.ExecModeReCheck {
+	if ctx.ExecMode() == sdk.ExecModeCheck || ctx.ExecMode() == sdk.ExecModeReCheck {
```
nivasan1, line 45: q: are there any metrics graphing the size of the tx-pool? Would be interested to see whether there is inconsistency / how large is it?

cosmos/runtime/txpool/handler.go
```diff
@@ -216,18 +216,18 @@ func (h *handler) broadcastTransaction(tx *ethtypes.Transaction, retries int) {
...
-	}
-
-	if rsp == nil || rsp.Code == 0 {
+	} else if rsp == nil {
 		return
 	}
 
 	switch rsp.Code {
+	case 0:
+	case 1:
```
nivasan1, line 229: nit: `rsp.Code` is a `DeliverTxCode`, no? If so, it'd be clearer to use the `DeliverTx` Code constants defined in comet

cosmos/runtime/txpool/handler.go
```diff
@@ -216,18 +216,18 @@ func (h *handler) broadcastTransaction(tx *ethtypes.Transaction, retries int) {
...
 		return
-	}
-
-	if rsp == nil || rsp.Code == 0 {
+	} else if rsp == nil {
 		return
 	}
 
 	switch rsp.Code {
+	case 0:
```
nivasan1, line 228: nit: why is this case necessary, looks like it j exits the switch?

cosmos/config/default.go
davidterpay: had a similar thought, does this gate txs from being added to a validator's mempool? is eip-1559 not the general min gas price for txs?

calbera: we dont actually check this min gas price on the cosmos side. All the gas logic is enforced by geth's txpool using 1559, etc

Issue #1481, opened 2024-02-04
feat: change README.md file to successfully install and use foundry


eminmtas:
`README.md` file is not clear about installing foundry. When we use `curl -L https://foundry.paradigm.xyz | bash` to install foundry, it keep gives this warning: 

```console
Detected your preferred shell is bash and added foundryup to PATH. Run 'source /home/codespace/.bashrc' or start a new terminal session to use foundryup.
Then, simply run 'foundryup' to install Foundry.
```
Screeenshot:
![image](https://github.com/berachain/polaris/assets/44838743/3fa356f3-3630-453d-b5d7-9862eaaa1196)

We just have to change the source and use `foundryup` command and then we can continue.

I think `README.md` file should be clearer. Just adding some lines of explanation will make it more clearer.

PR #1574, opened 2024-05-01
fix(readme): add foundryup command after curl

KhaledAylii:
Issue #1481

I ran into the same issue setting up, foundryup seems to always be needed to run after the curl

Description:
- Fix foundry install step


jcstein (approved):
LGTM, not sure how much weight my review carries though 😅

